Issue #1712, opened 2015-04-02
ASSERT on Linux: x64 tests/free shadow.c:261: false (fail to create shadow memory mapping)


derekbruening:
This happened once in a suite run, and I can't reproduce it:

```
drheapstat-dbg-32: all 30 tests passed
drheapstat-rel-32: all 30 tests passed
drmemory-dbg-32: all 67 tests passed
drmemory-dbg-64: 13 tests passed, **** 1 tests failed: ****
        free => ~~Dr.M~~ ASSERT FAILURE (thread 14198): /work/drmemory/git/src/drmemory/shadow.c:261: false (fail to create shadow memory mapping)

                CMake Error at runtest.cmake:311 (message):;  ~~Dr.M~~ ASSERT FAILURE (thread 14198):
                  /work/drmemory/git/src/drmemory/shadow.c:261: false (fail to create shadow
                  memory mapping)
drmemory-rel-32: all 67 tests passed
drmemory-rel-64: all 14 tests passed
final package: build successful; no tests for this build
```

derekbruening:
Happened on AppVeyor once too:

https://ci.appveyor.com/project/DynamoRIO/drmemory/build/1.0.17/job/j4j5j2o55j2b6n3w

derekbruening:
https://ci.appveyor.com/project/DynamoRIO/drmemory/build/1.0.30/job/du8ilypc06mckg4m

```
free.shadow => ~~Dr.M~~ ASSERT FAILURE (thread 2252): C:\projects\drmemory\drmemory\shadow.c:277: false (fail to create shadow memory mapping)
		Core dump file for application C:\projects\drmemory\build\build_drmemory-dbg-64\tests\free.exe (2108) created at C:\projects\drmemory\build\build_drmemory-dbg-64\logs\dynamorio\free.exe.2108.00000000.ldmp
		~~Dr.M~~ WARNING: application exited with abnormal code 0xffffffff;  ~~Dr.M~~ ASSERT FAILURE (thread 2252):
		  C:\projects\drmemory\drmemory\shadow.c:277: false (fail to create shadow
		  memory mapping)
```

derekbruening:
I can reproduce this in my win10 VM.  One problem is locale.nls which is #1810.  The other is a region with some app stacks.  So the Windows instances are a dup of #1810.  We'll leave this open for the Linux failures which have yet to reproduce.

derekbruening:
Happened in the loader test once on Travis:
[...]

PR #1950, opened 2017-03-12
DRi#2224: avoid asking for size from dr_safe_*

derekbruening:
Corrects uses of dr_safe_* that previously checked both the return value
and the size, as leaving the size parameter NULL is more efficient on
Windows 10 now due to kernel behavior changes.


derekbruening:
Appveyor failure is #1712

PR #1951, opened 2017-03-12
DRi#2192: pullall and review scripts

derekbruening:
Expands the pullall and review commands into scripts that give us these
features:

+ pullall pulls from both the upstream master and upstream feature branch,
  if it exists, using the appropriate rebase or merge in each case.

+ review checks for the branch name matching our conventions.

These are based on the DR scripts.


derekbruening:
wrap_wincrt and pcache-use failures are not on the list: will file issues for them.

PR #1962, opened 2017-03-15
i#1949: printing id of current thread for drltrace

### LINKED ISSUE #1949: Printing id of current thread that performing libcall for drltrace.
It shouldn't be hard to implement. There is already support for that in DynamoRIO.

mxmssh:
Added printing of current thread id and return address for every called libcall.

Fixes #1949

### DIFF
--- a/drltrace/drltrace.c
+++ b/drltrace/drltrace.c
@@ -173,8 +173,12 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
             dr_fprintf(outf, "<invalid memory>");
             /* Just keep going */
         });
+        /* XXX: we have to print a module id + offset here instead of an
+         * absolute address and provide a dump a module list at the end of
+         * execution to translate the module ids to paths.
+         */
         dr_fprintf(outf,
-                   (options.print_ret_addr == true) ? ") and return to "PFX: ")",
+                   options.print_ret_addr ? ") and return to " PFX: ")",
                    drwrap_get_retaddr(wrapcxt));
     }
     dr_fprintf(outf, "\n");
@@ -302,6 +306,7 @@ options_init(client_id_t id)
     dr_snprintf(options.logdir, BUFFER_SIZE_ELEMENTS(options.logdir), "-");
     options.all_args = 2;
     op_print_stderr = true;
+    /* XXX: we have to use droption here instead of manual parsing */
     for (s = dr_get_token(opstr, token, BUFFER_SIZE_ELEMENTS(token));
          s != NULL;
          s = dr_get_token(s, token, BUFFER_SIZE_ELEMENTS(token))) {
--- a/drltrace/drltrace.c
+++ b/drltrace/drltrace.c
@@ -161,7 +161,8 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
             dr_fprintf(outf, "<invalid memory>");
             /* Just keep going */
         });
-        dr_fprintf(outf, ")");
+        dr_fprintf(outf, ") thread id: "TIDFMT" and return to "PFX"",
+                   dr_get_thread_id(drcontext), drwrap_get_retaddr(wrapcxt));
     }
     dr_fprintf(outf, "\n");
     if (mod != NULL)
--- a/drltrace/drltrace.c
+++ b/drltrace/drltrace.c
@@ -82,6 +82,7 @@ static uint verbose;
 #define USAGE_CHECK(x, msg) DR_ASSERT_MSG(x, msg)
 
 #define OPTION_MAX_LENGTH MAXIMUM_PATH
+#define MAX_PREFIX 20
 
 typedef struct _drltrace_options_t {
     char logdir[MAXIMUM_PATH];
@@ -89,6 +90,7 @@ typedef struct _drltrace_options_t {
     bool ignore_underscore;
     char only_to_lib[MAXIMUM_PATH];
     uint all_args;
+    bool print_ret_addr;
 } drltrace_options_t;
 
 static drltrace_options_t options;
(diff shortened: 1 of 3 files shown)


drltrace/drltrace.c
```diff
@@ -161,7 +161,8 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
             dr_fprintf(outf, "<invalid memory>");
             /* Just keep going */
         });
-        dr_fprintf(outf, ")");
+        dr_fprintf(outf, ") thread id: "TIDFMT" and return to "PFX"",
```
derekbruening: How about using the thread id as a prefix, like DrM does, esp if this is going to stderr and then it would be part of the distinguish-from-app-output prefix.  That seems more natural, having it up front.

derekbruening: I have some concerns about the return address.  First, it seems like an optional feature not needed for all uses, so maybe it should be under a runtime option.  Second, everything else is symbolic: yet here is a raw address.  How does the user translate it to something symbolic without also having sthg like a drmodtrack modules log?

mxmssh: "How about using the thread id as a prefix, like DrM does"
Dou you mean this prefix:
~~Dr.M~~ (for example ~~2953~~) ?

mxmssh: "I have some concerns about the return address. First, it seems like an optional feature not needed for all uses, so maybe it should be under a runtime option."
Ok, will put it under optional feature.
"Second, everything else is symbolic: yet here is a raw address. How does the user translate it to something symbolic without also having sthg like a drmodtrack modules log?"

Well, from my practice of drltrace usage for malware analysis, it seems very important to understand actual address from where a function was called (I can later attach a debugger to check specific function, also I can understand that the same functions were called from different places which is also useful). I would prefer to stay it like that. It would be great if we can say that this address is a heap or previously allocated block of memory.

derekbruening: With ASLR (ignoring libs shared w/ other processes on Windows) moving libs around, the user may have no idea where the address is.  We should not assume it's easy to re-run the program, or require re-running and attaching a debugger on the 2nd run to make sense of the output.

It seems much more usable to do what drcov and drmemtrace use: if inside a module (99+% for non-DGC apps) print a module id + offset instead of an absolute address, and link in drmodtrack and dump a module list to translate the module ids to paths.

derekbruening: > "How about using the thread id as a prefix, like DrM does"
> Do you mean this prefix:
> ~~Dr.M~~ for example ~~2953~~

(There's a "quote" button to quote pasted text btw.)

Yes, a prefix at the start of each line identifying the thread.  Other tools use this format (e.g., Valgrind) so users may expect it.

mxmssh: I agree, will be a separate pull request.

drltrace/drltrace.c
```diff
@@ -161,7 +173,9 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
             dr_fprintf(outf, "<invalid memory>");
             /* Just keep going */
         });
-        dr_fprintf(outf, ")");
+        dr_fprintf(outf,
+                   (options.print_ret_addr == true) ? ") and return to "PFX: ")",
```
mxmssh: the string is too long but it looks better like that (only 1 character)

drltrace/drltrace.c
```diff
@@ -161,8 +173,9 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
             dr_fprintf(outf, "<invalid memory>");
             /* Just keep going */
         });
-        dr_fprintf(outf, ") thread id: "TIDFMT" and return to "PFX"",
-                   dr_get_thread_id(drcontext), drwrap_get_retaddr(wrapcxt));
+        dr_fprintf(outf,
+                   (options.print_ret_addr == true) ? ") and return to "PFX: ")",
```
derekbruening: No need for "== true"

derekbruening: This will fail to build on recent compilers (DRi#1959): needs a space " P

derekbruening: IMHO there should be a XXX comment about making the raw return address usable by providing module bounds

drltrace/drltrace.c
```diff
@@ -318,7 +331,9 @@ options_init(client_id_t id)
                 int res = dr_sscanf(token, "%u", &verbose);
                 USAGE_CHECK(res == 1, "invalid -verbose number");
             }
-        } else {
+        } else if (strcmp(token, "-print_ret_addr") == 0)
```
derekbruening: Should switch this to use droption (in a separate CL)

drltrace/drltrace.c
mxmssh: done

mxmssh: done

mxmssh: done

drltrace/drltrace.c
mxmssh: agree, put comment about that.

mxmssh:
A lot of failures, may be we can re-run the tests?

derekbruening:
> A lot of failures, may be we can re-run the tests?

Yes, but first please verify that nothing is new: file an issue if so, and if known, put a pointer to the issue number here in the PR.  I took a look and syscalls_win has never failed before: I filed #1974.

Unfortunately there's not a lot of manpower available to get Appveyor less flaky: any contributions to helping to fix the flaky tests is welcome.  In the past win7 and VS2010 were what we got down to near-zero flaky tests, but server12 and VS2013 on Appveyor have a bunch of work (still) to get to zero.

mxmssh:
Ok, I understand, will take a look when it finish.

PR #1963, opened 2017-03-15
i#1959: drltrace does incorrect parsing of logdir arg

### LINKED ISSUE #1959: drltrace_frontend does incorrect parsing of logdir input argument
There is a bug in a process of logdir parsing in drltrace.exe

mxmssh:
Fix a bug related to incorrect usage of NULL_TERMINATED_BUFFER on a pointer instead of string.

Fixes #1959

### DIFF
--- a/drltrace/drltrace_frontend.cpp
+++ b/drltrace/drltrace_frontend.cpp
@@ -250,7 +250,7 @@ check_logdir_path(char *logdir, size_t logdir_len) {
     else {
         dr_snprintf(logdir, logdir_len, "%s", absolute_logdir_path);
     }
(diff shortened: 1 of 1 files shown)


derekbruening:
For a minor issue in an in-progress project like drltrace that is not going to be noticed by others (drltrace in the drmemory repo has not been publicized and is not part of any release package) and that is going to be fixed immediately, there's a lot less value in filing an issue, so feel free to not file an issue: you can use iX-.  (If it's part of a released software package we might want a record of the problem in the tracker that users or other devs could search for; that's not the case here though.)

derekbruening:
nit: style violation: the title should have a colon after the number

PR #1965, opened 2017-03-19
Fix null character check in drmem_strndup()

maksqwe:
It may lead to out-of-bounds reading in case when "strlen(src) + 1 < max"


derekbruening:
The "unknown repository" here precluded using the update-branch button so committed manually in 460a4a877f68f19551eb91713b9a9c9857dc05f3

Issue #1929, opened 2016-12-12
ASSERT nonaslr test failing with reachability assert in DR


derekbruening:
http://dynamorio.org/CDash/testDetails.php?test=167634&build=15892

```
<Starting application /work/drmemory/nightly/run/build_drmemory-dbg-64/tests/nonaslr (7972)>
<Paste into GDB to debug DynamoRIO clients:
set confirm off
add-symbol-file '/work/drmemory/nightly/run/build_drmemory-dbg-64/bin64/debug/libdrmemorylib.so' 0x000000007381e000
add-symbol-file '/work/drmemory/nightly/run/build_drmemory-dbg-64/dynamorio/lib64/debug/libdynamorio.so' 0x000000007102d808
add-symbol-file '/usr/lib64/libstdc++.so.6' 0x00007f65c7196fa0
add-symbol-file '/usr/lib64/libm.so.6' 0x00007f65c6e09550
add-symbol-file '/usr/lib64/libc.so.6' 0x00007f65c6a614d0
add-symbol-file '/usr/lib64/ld-linux-x86-64.so.2' 0x00007f65c681ead0
add-symbol-file '/usr/lib64/libgcc_s.so.1' 0x00007f65c6608af0
>
[...]

Issue #1741, opened 2015-07-20
wrap_wincrt test fails in all Windows builds


byron-hawkins:
The first intended error does not occur in `wincrt.cpp` for any of the Windows builds. The erroneous statement over-reads the 8-byte allocation `foo` by 40 bytes. Looking at the heap in windbg, it appears that there is a different heap at 56 bytes above `foo`, but this would not include the over-read, so it does not explain the lack of an error.

derekbruening:
Something about your machine results in a pre-existing (at our late injection point) alloc already there when it's not on other machines.  But basically the test is flaky.  Originally all of those accesses were trying to test the crtdbg redzones, and I think 48 was the size we had to exceed.
Probably such an access outside of the redzone should just be removed from the test.

derekbruening:
This has started failing on Appveyor every once in a while:

https://ci.appveyor.com/project/DynamoRIO/drmemory/build/1.0.65/job/3033i1n94qyvkn8g

```
wrap_wincrt =>   stderr failed to match "~~Dr\.M~~ 5 unique, 5 total unaddressable
		  access\(es\)", found "~~Dr.M~~ 4 unique, 4 total unaddressable access(es)"
		  instead
```

derekbruening:
This is now failing very frequently on Appveyor:

https://ci.appveyor.com/project/DynamoRIO/drmemory/build/1.0.164

PR #1966, opened 2017-03-19
Fix malloc_entry_t->struct_size initialization

maksqwe:


derekbruening:
Hmm, the Travis failure is in the nonaslr test and is #1929.  That had not reproduced in a long time.

derekbruening:
Appveyor failure is #1741.  Grr, these new CI setups need more work.

PR #1975, opened 2017-04-07
i#1948 add libcall args sym printing: syscall args support

mxmssh:
Added arguments printing of already known syscalls from drsyscall's
tables for Windows and Linux.

### DIFF
--- a/drltrace/drltrace.c
+++ b/drltrace/drltrace.c
@@ -62,8 +62,8 @@
  * + Add argument values and return values.  The number and type of each
  *   argument and return would likely come from the filter configuration
  *   file, or from querying debug information.
- *   Today we have simple type-blind printing via -all_args and usage of
- *   drsyscall to print symbolic arguments for known library calls.
+ *   Today we have simple type-blind printing via -num_unknown_args and
+ *   usage of drsyscall to print symbolic arguments for known library calls.
  *
  * + Add 2 more modes, both gathering statistics rather than a full
  *   trace: one mode that counts total calls, and one that just
@@ -89,7 +89,7 @@ typedef struct _drltrace_options_t {
     bool only_from_app;
     bool ignore_underscore;
     char only_to_lib[MAXIMUM_PATH];
-    uint unknown_args;
+    uint num_unknown_args;
     uint max_args;
     bool print_ret_addr;
 } drltrace_options_t;
@@ -105,9 +105,7 @@ static app_pc exe_start;
 /* runtest.cmake assumes this is the prefix, so update both when changing it */
 #define STDERR_PREFIX "~~~~ "
 
-#define MAX_LIBCALL_NAME 255
-
-/* XXX: The functions print_simple_value and print_arg were taken from drstrace.
+/* XXX i#1978: The functions print_simple_value and print_arg were taken from drstrace.
  * It would be better to move them in drsyscall and import in drstrace and here.
  */
 static void
@@ -176,17 +174,17 @@ print_args_unknown_call(app_pc func, void *wrapcxt)
 {
     uint i;
     void *drcontext = drwrap_get_drcontext(wrapcxt);
-    dr_fprintf(outf, "(");
     DR_TRY_EXCEPT(drcontext, {
-        for (i = 0; i < options.unknown_args; i++) {
-            dr_fprintf(outf, "%s" PFX, (i != 0) ? ", " : "",
+        for (i = 0; i < options.num_unknown_args; i++) {
+            dr_fprintf(outf, "\n    arg %d: " PFX, i,
                        drwrap_get_arg(wrapcxt, i));
         }
     }, {
         dr_fprintf(outf, "<invalid memory>");
         /* Just keep going */
     });
-    dr_fprintf(outf, ")");
+    /* all args have been sucessfully printed */
+    dr_fprintf(outf, options.print_ret_addr ? "\n   ": "");
 }
 
 static void
@@ -194,27 +192,22 @@ print_symbolic_args(const char *name, void *wrapcxt, app_pc func)
 {
     drmf_status_t res;
     drsys_syscall_t *syscall;
-    char syscall_name[MAX_LIBCALL_NAME];
 
-    if (options.max_args > 0) {
-        if (name[0] == 'Z' && name[1] == 'w')
-            dr_snprintf(syscall_name, BUFFER_SIZE_ELEMENTS(syscall_name), "Nt%s", name+2);
-        else
-            dr_snprintf(syscall_name, BUFFER_SIZE_ELEMENTS(syscall_name), "%s", name);
+    if (options.max_args == 0)
+        return;
 
-        res = drsys_name_to_syscall(syscall_name, &syscall);
-        if (res == DRMF_SUCCESS) {
-            res = drsys_iterate_arg_types(syscall, drlib_iter_arg_cb, wrapcxt);
-            if (res != DRMF_SUCCESS && res != DRMF_ERROR_DETAILS_UNKNOWN)
-                ASSERT(false, "drsys_iterate_arg_types failed in print_symbolic_args");
-            /* all args have been sucessfully printed */
-            dr_fprintf(outf, options.print_ret_addr ? "\n   ": "");
-            return;
-        } else {
-            /* use standard type-blind scheme */
-            if (options.unknown_args > 0)
-                print_args_unknown_call(func, wrapcxt);
-        }
+    res = drsys_name_to_syscall(name, &syscall);
+    if (res == DRMF_SUCCESS) {
+        res = drsys_iterate_arg_types(syscall, drlib_iter_arg_cb, wrapcxt);
+        if (res != DRMF_SUCCESS && res != DRMF_ERROR_DETAILS_UNKNOWN)
+            ASSERT(false, "drsys_iterate_arg_types failed in print_symbolic_args");
+        /* all args have been sucessfully printed */
+        dr_fprintf(outf, options.print_ret_addr ? "\n   ": "");
+        return;
+    } else {
+        /* use standard type-blind scheme */
+        if (options.num_unknown_args > 0)
+            print_args_unknown_call(func, wrapcxt);
     }
 }
 
@@ -277,12 +270,12 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
 
     /* XXX: We employ two schemes of arguments printing. drsyscall is used
      * to get a symbolic representation of arguments for known library calls.
-     * For the rest of library calls we use type-blind printing and -unknown_args
+     * For the rest of library calls we use type-blind printing and -num_unknown_args
      * to get a count of arguments to print.
      */
     print_symbolic_args(name, wrapcxt, func);
 
-    /* XXX: we have to print a module id + offset here instead of an
+    /* XXX i#1979: we have to print a module id + offset here instead of an
      * absolute address and provide a list of modules at the end of
      * execution to translate the module ids to paths.
      */
@@ -416,7 +409,7 @@ options_init(client_id_t id)
     /* default values */
     dr_snprintf(options.logdir, BUFFER_SIZE_ELEMENTS(options.logdir), "-");
     options.max_args = 6;
-    options.unknown_args = 2;
+    options.num_unknown_args = 2;
     op_print_stderr = true;
     /* XXX: we have to use droption here instead of manual parsing */
     for (s = dr_get_token(opstr, token, BUFFER_SIZE_ELEMENTS(token));
@@ -438,8 +431,8 @@ options_init(client_id_t id)
             s = dr_get_token(s, token, BUFFER_SIZE_ELEMENTS(token));
             USAGE_CHECK(s != NULL, "missing -num_unknown_args number");
             if (s != NULL) {
-                int res = dr_sscanf(token, "%u", &options.unknown_args);
-                USAGE_CHECK(res == 1, "invalid -all_args number");
+                int res = dr_sscanf(token, "%u", &options.num_unknown_args);
+                USAGE_CHECK(res == 1, "invalid -num_unknown_args number");
             }
         } else if (strcmp(token, "-num_max_args") == 0) {
             s = dr_get_token(s, token, BUFFER_SIZE_ELEMENTS(token));
@@ -448,8 +441,7 @@ options_init(client_id_t id)
                 int res = dr_sscanf(token, "%u", &options.max_args);
                 USAGE_CHECK(res == 1, "invalid -num_max_args number");
             }
-        }
-        else if (strcmp(token, "-verbose") == 0) {
+        } else if (strcmp(token, "-verbose") == 0) {
             s = dr_get_token(s, token, BUFFER_SIZE_ELEMENTS(token));
             USAGE_CHECK(s != NULL, "missing -verbose number");
             if (s != NULL) {
--- a/drltrace/drltrace.c
+++ b/drltrace/drltrace.c
@@ -51,6 +51,7 @@
 #include "drwrap.h"
 #include "drx.h"
 #include "utils.h"
+#include "drsyscall_os.h"
 #include <string.h>
 
 /* XXX i#1349: features to add:
@@ -62,7 +63,8 @@
  * + Add argument values and return values.  The number and type of each
  *   argument and return would likely come from the filter configuration
  *   file, or from querying debug information.
- *   Today we have simple type-blind printing via -all_args.
+ *   Today we have simple type-blind printing via -all_args and usage of
+ *   drsyscall to print symbolic arguments for known library calls.
  *
  * + Add 2 more modes, both gathering statistics rather than a full
  *   trace: one mode that counts total calls, and one that just
@@ -82,7 +84,6 @@ static uint verbose;
 #define USAGE_CHECK(x, msg) DR_ASSERT_MSG(x, msg)
 
 #define OPTION_MAX_LENGTH MAXIMUM_PATH
-#define MAX_PREFIX 20
 
 typedef struct _drltrace_options_t {
     char logdir[MAXIMUM_PATH];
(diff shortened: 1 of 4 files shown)


mxmssh:
I am planning to use the same syscall_info_t structure to describe other libcalls.

drltrace/drltrace_frontend.cpp
```diff
@@ -97,6 +97,12 @@ static droption_t<bool> op_print_ret_addr
 (DROPTION_SCOPE_CLIENT, "print_ret_addr", false, "Print library call's return address",
  "Print return addresses of library calls.");
 
+static droption_t<uint> op_all_args
+(DROPTION_SCOPE_CLIENT, "all_args", 2, "Number of library call arguments to print",
+ "Number of library call arguments to print. Specify 0 to disable. Drltrace will print "
+ "symbolic representation of arguments in case of known library call including all "
+ "known arguments.");
```
derekbruening: The description is confusing: so if the actual arguments are known, they will all be printed and this option ignored, right?  So then this should option should be named "num_unknown_args" or something, right?  And the short description should have "unknown" in it.

drltrace/drltrace.c
```diff
@@ -160,33 +289,116 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
...
+    for (i = 0; i < num_user32_syscalls(); i++)
+        hashtable_add_new_name_entry(syscall_user32_info[i], &syscall_user32_info[i]);
+    for (i = 0; i < num_gdi32_syscalls(); i++)
+        hashtable_add_new_name_entry(syscall_gdi32_info[i], &syscall_gdi32_info[i]);
+    for (i = 0; i < num_ntdll_syscalls(); i++)
+        hashtable_add_new_name_entry(syscall_ntdll_info[i], &syscall_ntdll_info[i]);
+#elif MACOS
+    for (i = 0; i < count_syscall_info_bsd; i++)
+        hashtable_add_new_name_entry(syscall_info_bsd[i], &syscall_info_bsd[i]);
+#else /* LINUX */
```
derekbruening: Safer to have #elif defined(LINUX) and #else #error

drltrace/drltrace.c
```diff
@@ -160,33 +289,116 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
...
+#ifdef WINDOWS
+    for (i = 0; i < num_kernel32_syscalls(); i++)
+        hashtable_add_new_name_entry(syscall_kernel32_info[i], &syscall_kernel32_info[i]);
+    for (i = 0; i < num_user32_syscalls(); i++)
+        hashtable_add_new_name_entry(syscall_user32_info[i], &syscall_user32_info[i]);
+    for (i = 0; i < num_gdi32_syscalls(); i++)
+        hashtable_add_new_name_entry(syscall_gdi32_info[i], &syscall_gdi32_info[i]);
+    for (i = 0; i < num_ntdll_syscalls(); i++)
+        hashtable_add_new_name_entry(syscall_ntdll_info[i], &syscall_ntdll_info[i]);
+#elif MACOS
```
derekbruening: defined(MACOS)

drltrace/drltrace.c
```diff
@@ -160,33 +289,116 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
...
+     * to get a symbolic representation of arguments for known library calls.
+     * For the rest of library calls as well as for other OSs we use type-blind
+     * printing using -all_args to get a count of arguments to print. It would be
+     * great to have the one scheme for all supported OSs.
+     */
+    if (options.all_args > 0)
+        print_symbolic_args(func, wrapcxt);
+
+    /* XXX: we have to print a module id + offset here instead of an
+     * absolute address and provide a dump a module list at the end of
```
derekbruening: "provide a dump a"

drltrace/drltrace.c
```diff
@@ -104,6 +105,134 @@ static app_pc exe_start;
...
+                       drwrap_get_arg(wrapcxt, i));
+        }
+    }, {
+        dr_fprintf(outf, "<invalid memory>");
+        /* Just keep going */
+    });
+    dr_fprintf(outf, ")");
+}
+
+void print_symbolic_args(app_pc func, void *wrapcxt)
```
derekbruening: style violation again

derekbruening: static

drltrace/drltrace.c
```diff
@@ -104,6 +105,134 @@ static app_pc exe_start;
...
+    print_arg(arg);
+    return true; /* keep going */
+}
+
+syscall_info_t *lookup_entry(app_pc func)
+{
+    return (syscall_info_t *)hashtable_lookup(&addr2entry_table, func);
+}
+
+void print_args(app_pc func, void *wrapcxt)
```
derekbruening: style violation: type on own line

derekbruening: static

drltrace/drltrace.c
```diff
@@ -104,6 +105,134 @@ static app_pc exe_start;
...
+{
+    if (arg->ordinal == -1)
+        return true;
+    arg->value = (ptr_uint_t)drwrap_get_arg(wrapcxt, arg->ordinal);
+
+    print_arg(arg);
+    return true; /* keep going */
+}
+
+syscall_info_t *lookup_entry(app_pc func)
```
derekbruening: style violation

derekbruening: static

drltrace/drltrace.c
```diff
@@ -104,6 +105,134 @@ static app_pc exe_start;
...
+}
+
+void print_symbolic_args(app_pc func, void *wrapcxt)
+{
+    drmf_status_t res;
+    syscall_info_t *libcall_entry;
+    libcall_entry = lookup_entry(func);
+    if (libcall_entry != NULL) {
+        /* drsys_iterate_arg_types employs static enumeration. */
+        res = drsys_iterate_arg_types((drsys_syscall_t *)libcall_entry,
```
derekbruening: This breaks the abstraction barrier of the drsyscall library: drsys_syscall_t is an opaque type and a user of the library should not have assumptions about what it looks like.  See the suggestion above for using library interfaces instead.

drltrace/drltrace.c
```diff
@@ -104,6 +105,134 @@ static app_pc exe_start;
...
+#define NAME2ENTRY_TABLE_HASH_BITS 13 /* 1.5K of them, x2 for no-prefix entries + Zw */
+static hashtable_t name2entry;
+
+/* Table that maps address (found when dll is loading) to the syscall_info_t structure */
+#define ADDR2ENTRY_TABLE_HASH_BITS 8
+static hashtable_t addr2entry_table;
+
+/* XXX: We need to add support of printing from secondary table (e.g. ioctl) (xref i#1549) */
+#ifdef WINDOWS
+extern dr_os_version_info_t win_ver; /* already defined in drsyscall_windows.c */
```
derekbruening: None of these are exported by the drsyscall library, though.  They're not part of the library interface!  All uses here should be through interfaces.

It sounds like you could add a static drsys_lookup_syscall() routine that takes in a name and returns a drsys_syscall_t "handle", right?  Then there's no need to break the library's abstractions in multiple ways here by using its internal, private data.

drltrace/drltrace.c
mxmssh: done

drltrace/drltrace.c
mxmssh: done

drltrace/drltrace.c
mxmssh: done

drltrace/drltrace.c
mxmssh: Agree, implemented a new function: drsys_iterate_arg_types_by_sysname

drltrace/drltrace.c
mxmssh: done

mxmssh: done

drltrace/drltrace.c
mxmssh: done

mxmssh: done

drltrace/drltrace.c
mxmssh: done

drltrace/drltrace.c
mxmssh: Agree, I've created three functions that user can use externally.

drltrace/drltrace_frontend.cpp
mxmssh: Well, the idea was simple: if user wants to print arguments, he would use -all_args and in this case drltrace will provide as much as possible information. If user doesn't want to print he can specify 0 and disable arguments printing at all (for example to increase performance).

drltrace/drltrace.c
derekbruening: IMHO it's cleaner to separate things and have a lookup whose output can be used to the existing iterator, rather than having a multi-operation iterator.

BTW, if drsys_name_to_syscall() can be used statically, which should be true, then no new interface function is needed, right?  Call drsys_name_to_syscall() and pass the resulting handle to drsys_iterate_arg_types().

drltrace/drltrace_frontend.cpp
derekbruening: I would suggest splitting into two options: one that controls the number of unknown args printed ("num_unknown _args" or sthg), and another that is either a max on args of any type ("max_args" or sthg) or maybe just a boolean controlling whether any arg printing is done (the max seems more useful).

drltrace/drltrace.c
mxmssh: It makes sense, what about hashtables initialization ? we have drsyscall_os_init but I need to provide a context for this function...

derekbruening: I'm not sure what you mean: any hashtables inside drsyscall are invisible to the user here and are initialized when drsyscall_init() is called.

mxmssh: Yes, drsys_init()/drsys_exit() will do everything. So it works perfect the only one problem is ASSERT at lines 252-257, it raises error when I am sending Zw* syscall. I can replace Zw* with Nt* but probably I am missing something ?

drltrace/drltrace_frontend.cpp
mxmssh: I am not sure that I understand. For example user specified -num_unknown_args 3 -max_args 7.
Should we print 3 arguments for unknown library calls and 7 arguments for known libcalls or 7 arguments for both types?
What should we have by default for -max_args?
Boolean controlling looks more reasonable for me. For example user can specify -num_unknown_args 5 -print_known_args. So we know that we should print everything for known system calls and maximum 5 arguments for unknown. By default we may set -num_unknown_args to be 0 and -print_known_args to be true ? (or may be false to demonstrate maximum performance by default).

drltrace/drltrace.c
mxmssh: ASSERT a lines 252-257 in drsyscall.c

drltrace/drltrace_frontend.cpp
derekbruening: "-num_unknown_args 3 -max_args 7" means *never* print more than 7, and furthermore don't print more than 3 when the types and # of args are not known.

Definitely there is a use case where no args should be printed.  The question is whether some users would want to limit the # of args to some small number for cleaner output: see some but not all.  On Linux there are never more than 6, but on Windows where there can be over 20 it may be a different story.

mxmssh: Ok, I got it, will implement it this way

drltrace/drltrace.c
```diff
@@ -62,7 +62,8 @@
  * + Add argument values and return values.  The number and type of each
  *   argument and return would likely come from the filter configuration
  *   file, or from querying debug information.
- *   Today we have simple type-blind printing via -all_args.
+ *   Today we have simple type-blind printing via -all_args and usage of
```
derekbruening: num_unknown_args

drltrace/drltrace.c
```diff
@@ -104,6 +105,119 @@ static app_pc exe_start;
 /* runtest.cmake assumes this is the prefix, so update both when changing it */
 #define STDERR_PREFIX "~~~~ "
 
+#define MAX_LIBCALL_NAME 255
+
+/* XXX: The functions print_simple_value and print_arg were taken from drstrace.
+ * It would be better to move them in drsyscall and import in drstrace and here.
```
derekbruening: Yes, this is a lot of code to duplicate.  Is there an issue # for the XXX?

drltrace/drltrace.c
```diff
@@ -104,6 +105,119 @@ static app_pc exe_start;
...
+        /* We assume little-endian */
+        if (dr_safe_read((void *)arg->value, arg->size, &deref, NULL))
+            dr_fprintf(outf, (leading_zeroes ? " => " PFX : " => " PIFX), deref);
+    }
+}
+
+static void
+print_arg(drsys_arg_t *arg)
+{
+    dr_fprintf(outf, "\n    arg %d: ", arg->ordinal);
```
derekbruening: This format is completely different from the unknown-arg format: the unknown is all one one line, separated by commas.  This has a separate line per arg.  Shouldn't they be consistent?

drltrace/drltrace.c
```diff
@@ -104,6 +105,119 @@ static app_pc exe_start;
...
+{
+    drmf_status_t res;
+    drsys_syscall_t *syscall;
+    char syscall_name[MAX_LIBCALL_NAME];
+
+    if (options.max_args > 0) {
+        if (name[0] == 'Z' && name[1] == 'w')
+            dr_snprintf(syscall_name, BUFFER_SIZE_ELEMENTS(syscall_name), "Nt%s", name+2);
+        else
+            dr_snprintf(syscall_name, BUFFER_SIZE_ELEMENTS(syscall_name), "%s", name);
```
derekbruening: We could avoid this redundant copy by having a pointer that points either at name or syscall_name

drltrace/drltrace.c
```diff
@@ -104,6 +105,119 @@ static app_pc exe_start;
...
+
+        res = drsys_name_to_syscall(syscall_name, &syscall);
```
derekbruening: Or should we change this routine to succeed with the Zw variant?

drltrace/drltrace.c
```diff
@@ -160,27 +274,21 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
...
-         * execution to translate the module ids to paths.
-         */
-        dr_fprintf(outf,
-                   options.print_ret_addr ? ") and return to " PFX: ")",
-                   drwrap_get_retaddr(wrapcxt));
-    }
+
+    /* XXX: We employ two schemes of arguments printing. drsyscall is used
+     * to get a symbolic representation of arguments for known library calls.
+     * For the rest of library calls we use type-blind printing and -unknown_args
```
derekbruening: num_

drltrace/drltrace.c
```diff
@@ -160,27 +274,21 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
...
+     * to get a count of arguments to print.
+     */
+    print_symbolic_args(name, wrapcxt, func);
+
+    /* XXX: we have to print a module id + offset here instead of an
```
derekbruening: i#?

drltrace/drltrace.c
```diff
@@ -104,6 +105,119 @@ static app_pc exe_start;
...
+}
+
+static void
+print_symbolic_args(const char *name, void *wrapcxt, app_pc func)
+{
+    drmf_status_t res;
+    drsys_syscall_t *syscall;
+    char syscall_name[MAX_LIBCALL_NAME];
+
+    if (options.max_args > 0) {
```
derekbruening: suggest: "if (options.max_args == 0) return;" to make the rest of the function un-indented.

drltrace/drltrace_frontend.cpp
```diff
@@ -97,6 +97,16 @@ static droption_t<bool> op_print_ret_addr
 (DROPTION_SCOPE_CLIENT, "print_ret_addr", false, "Print library call's return address",
  "Print return addresses of library calls.");
 
+static droption_t<uint> op_unknown_args
+(DROPTION_SCOPE_CLIENT, "num_unknown_args", 2, "Number of unknown libcall args to print",
+ "Number of arguments to print for unknown library calls.  Specify 0 to disable "
+ "unknown args printing.");
+
+static droption_t<uint> op_max_args
+(DROPTION_SCOPE_CLIENT, "num_max_args", 6, "Maximum number of arguments to print",
```
derekbruening: It's not ideal to duplicated these default values.  The plan is to switch drltrace.c to use droption also and share these definitions directly, right?

drltrace/drltrace.c
mxmssh: So, do we want to print unknown args one per line?

drltrace/drltrace_frontend.cpp
mxmssh: Yes, I'm preparing a sperate commit for that.

drltrace/drltrace.c
mxmssh: Do we want additional flag e.g. ignore_zw_nt_mismatch or sthg like that ?

drltrace/drltrace.c
mxmssh: done

drltrace/drltrace.c
mxmssh: done

drltrace/drltrace.c
mxmssh: Done

drltrace/drltrace.c
mxmssh: Done

drltrace/drltrace.c
mxmssh: Created

drltrace/drltrace.c
derekbruening: I thought drsyscall had entries for both the Nt and Zw forms.  Are you sure it doesn't?  I thought that today either query would work.

derekbruening: I see code in drsyscall_windows.c that is adding duplicate entries for Zw to the name2num_table.

mxmssh: this is correct both versions work. We have an assert at https://github.com/DynamoRIO/drmemory/blob/master/drsyscall/drsyscall.c#L251

mxmssh: NtQueryPerformanceCounter ZwQueryPerformanceCounter
ASSERT FAILURE (thread 2760): C:\drmemory\drmemory\drsyscall\drsyscall.c:257: st
ri_eq(sysinfo->name, name) IF_WINDOWS(|| strcasestr(sysinfo->name, name) != NULL
) (name<->num mismatch)

derekbruening: The assert should be relaxed of course.

drltrace/drltrace.c
mxmssh: I am worry about log size here, compact unknown args printing scheme saves a lot of space...

derekbruening: If you want compact, should the known args be one line (x, y, z, ...) then, closer to Linux ltrace and strace?

mxmssh: In case of known I think we have pretty good scheme of printing. One line would be a bit complicated to read in case of known arguments while for unknown arguments it looks ok. I would prefer to leave current scheme for now...

derekbruening: That sounds like log size or screen space is not a concern, then.

Having different formats complicates post-processing the log: now to parse the file there are two different formats to match.

It also seems odd to have a syscall with a known name but unknown types (e.g., NtCreateEnclave) have its 9 arguments all on separate lines with unknown types, while a syscall with an unknown name is printed completely differently even though its arg types are also unknown.

mxmssh: makes sense, ok let's do one arg per line for all schemes.

drsyscall/drsyscall.c
```diff
@@ -247,13 +250,21 @@ drsys_name_to_syscall(const char *name, drsys_syscall_t **syscall OUT)
         ASSERT(false, "name2num should return num in systable");
         return DRMF_ERROR_NOT_FOUND;
     }
+
 #ifdef DEBUG
-    ASSERT(stri_eq(sysinfo->name, name)
+    /* ignore possible Nt/Zw mismatch */
+    if (((sysinfo->name[0] == 'N' && sysinfo->name[1] == 't') ||
```
derekbruening: Should probably be Windows only

derekbruening:
There's no test of the output yet?  We should add a test soon.

drsyscall/drsyscall.c
mxmssh: Will change in the next commit.

mxmssh:
Yes, it's important.

drsyscall/drsyscall.c
derekbruening: Why the next commit?  It needs to update to master and thus re-run the checks anyway, right?

derekbruening: (I'm assuming by "next commit" you mean tacking onto a future pull request?  Seems cleaner to be part of this one, not an unrelated pull request, right?)

mxmssh: yes, yes, just need to pass through the local tests.

PR #1980, opened 2017-04-14
i#1979: print module id + offset for ret addr in drltrace

### LINKED ISSUE #1979: Print module id + offset for return address in drltrace
We have to print a module id + offset for return addresses in drltrace instead of an absolute address which we use now and provide a list of modules at the end of execution to translate the module ids to paths.

mxmssh:
Implemented module id + offset printing scheme instead of
raw address for return addresses in drltrace.

Fixes #1979

### DIFF
--- a/drltrace/drltrace.c
+++ b/drltrace/drltrace.c
@@ -284,7 +284,7 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
         res = drmodtrack_lookup(drcontext, ret_addr, &mod_id, &mod_start);
         if (res == DRCOVLIB_SUCCESS) {
             dr_fprintf(outf,
-                       options.print_ret_addr ? " and return to module id:%d, offset:" PIFX: "",
+                       options.print_ret_addr ? " and return to module id:%d, offset:" PIFX : "",
                        mod_id, ret_addr - mod_start);
         }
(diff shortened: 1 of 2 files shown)


drltrace/drltrace.c
```diff
@@ -275,13 +279,15 @@ lib_entry(void *wrapcxt, INOUT void **user_data)
...
-     */
-    dr_fprintf(outf,
-               options.print_ret_addr ? " and return to " PFX: "",
-               drwrap_get_retaddr(wrapcxt));
+    if (options.print_ret_addr) {
+        ret_addr = drwrap_get_retaddr(wrapcxt);
+        res = drmodtrack_lookup(drcontext, ret_addr, &mod_id, &mod_start);
+        if (res == DRCOVLIB_SUCCESS) {
+            dr_fprintf(outf,
+                       options.print_ret_addr ? " and return to module id:%d, offset:" PIFX: "",
```
derekbruening: suggest: space before : part of ?: to make it easier to read

mxmssh: Do you mean space after PIFX ?

derekbruening: Yes, I found it hard to parse as written.

PR #1989, opened 2017-05-12
i#1948 args printing for libcalls: initial commit

mxmssh:
Added library calls printing in drysyscall. Library
calls are described using the same scheme as for syscalls.
Added a new option for drsys_init to be able to use libcalls.
Added CreateFileA/CreateFileW libcalls to test new scheme.


derekbruening:
Hmm I guess I'm not sure about adding library databases inside drsyscall.  See comments in the issue.

mxmssh:
I'm closing this pull request since we've decided to go another way.

Issue #1938, opened 2017-02-19
set up AppVeyor and new feature-branch-based workflow


derekbruening:
We want to add Windows CI testing via AppVeyor, just like for DR: https://github.com/DynamoRIO/dynamorio/issues/2145

Xref Travis CI: https://github.com/DynamoRIO/dynamorio/issues/1962

Once that's in place we want to switch DrMem to use a feature-branch-based workflow with pull requests, like we've done for DR.

derekbruening:
MinGW is failing for unknown reasons.  No error message is printed: it just exits with code 1.

https://ci.appveyor.com/project/DynamoRIO/drmemory/build/1.0.7/job/hv0w1uom29y5wisq

```
-- Found MinGW gcc at C:/MinGW/bin/mingw32-g++.exe
-- Found C:/MinGW/bin/libstdc++-6.dll, enabling shared lib MinGW test

[113/823] Generating cs2bugMinGW_shared.exe
FAILED: tests/cs2bugMinGW_shared.exe 
cmd.exe /C "cd /D C:\projects\drmemory\build\build_drmemory-dbg-32\tests && "C:\Program Files (x86)\CMake\bin\cmake.exe" -D exename=C:/projects/drmemory/build/build_drmemory-dbg-32/tests/cs2bugMinGW_shared.exe -D gcc=C:/MinGW/bin/mingw32-g++.exe -D source=C:/projects/drmemory/tests/cs2bug.cpp -D args= -P C:/projects/drmemory/tests/rungcc.cmake"
CMake Error at C:/projects/drmemory/tests/rungcc.cmake:74 (message):
  *** C:/MinGW/bin/mingw32-g++.exe failed (1): ***
```

Custom diagnostics:
https://ci.appveyor.com/project/DynamoRIO/drmemory/build/job/etuidso68qypx6ea

```
# C:\MinGW\bin\mingw32-g++.exe -v
Using built-in specs.
COLLECT_GCC=C:\MinGW\bin\mingw32-g++.exe
COLLECT_LTO_WRAPPER=c:/mingw/bin/../libexec/gcc/mingw32/4.9.3/lto-wrapper.exe
Target: mingw32
[...]

PR #1998, opened 2017-05-30
DRi#2453: add XP support for VS2013 builds

derekbruening:
Updates DR to 266b8a3 to fix DRi#2453 to support VS2013 client builds on XP.


derekbruening:
appveyor failure is pcache-use under #1938

PR #1999, opened 2017-06-02
Umbra docs: Adds note to umbra_insert_app_to_shadow

toshipiazza:
`umbra_insert_app_to_shadow()` destroys aflags. We add a reminder in the
documentation to save and restore aflags before we call this method.

### DIFF
--- a/umbra/umbra.h
+++ b/umbra/umbra.h
@@ -389,6 +389,9 @@ DR_EXPORT
  * \note: \p num_scratch_regs must not be smaller than the value returned from
  * umbra_num_scratch_regs_for_translation, otherwise error code
  * DRMF_ERROR_NOT_ENOUGH_REGS is returned.
+ *
+ * \note: This method destroys aflags. Be sure to save and restore aflags before
+ * and after this method is called, i.e. with \p drreg_reserve_aflags.
  */
 drmf_status_t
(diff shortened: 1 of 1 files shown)


umbra/umbra.h
```diff
@@ -389,6 +389,9 @@ DR_EXPORT
  * \note: \p num_scratch_regs must not be smaller than the value returned from
  * umbra_num_scratch_regs_for_translation, otherwise error code
  * DRMF_ERROR_NOT_ENOUGH_REGS is returned.
+ *
+ * \note: This method destroys aflags. Be sure to save and restore aflags before
+ * and after this method is called, i.e. with \p drreg_reserve_aflags.
```
derekbruening: s/i.e./e.g.,/

Also if you put () after the function name doxygen will auto-link it (though prob won't work from DrM to DR since the docs are built separately)

toshipiazza: Done

toshipiazza:
What is the protocol for handling failed tests in the test suite (looks unrelated)? I don't see an issue up for the `bitfield` test crash, so should I create an issue?

derekbruening:
Yes please file an issue if it doesn't have its own or isn't listed in the appveyor issue.  I can merge this.

derekbruening:
Note that the branch name is supposed to be sthg like iX-umbra-docs

PR #2001, opened 2017-06-14
i#1948 drltrace libcall args printing: config file init

mxmssh:
Implemented simplified language for function prototypes description
to provide an external source where user can define custom libcalls
to be able to print arguments.

Added config file parsing routines
Added hashtable to store libcalls info
Added strings printing routine
Added a third way to print arguments
Added simplified config file with strcmp/wcscmp/printf/puts prototypes
Added tests for Windows and Linux


drltrace/drltrace.cpp
```diff
@@ -98,6 +369,15 @@ print_simple_value(drsys_arg_t *arg, bool leading_zeroes)
 }
 
 static void
+print_string(void *pointer_str, bool is_wide)
+{
+    if (pointer_str == NULL)
+        dr_fprintf(outf, "<null>");
+    else
+        dr_fprintf(outf, is_wide ? "%S" : "%s", pointer_str);
```
mxmssh: do we have a safe way to print strings from memory?

derekbruening:
FYI I'm traveling and may not get to this until next week

drltrace/drltrace.cpp
derekbruening: There's no existing convenience routine: you'd have to use a DR try-except or write a loop that uses a safe read.

mxmssh: ok, will fix it after review in the next commit.

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+
+    if (dr_file_size(file_desc, &size_to_read) == false) {
+        VNOTIFY(0, "unable to open config file at %s, config is not used" NL,
+                    op_config_file.get_value().c_str());
+        op_disable_config.set_value(true);
+        dr_close_file(file_desc);
+        return;
+    }
+
+    buf = (char *)global_alloc(size_to_read, HEAPSTAT_MISC);
```
derekbruening: Generally our approach to parsing text files in existing code (drsymcache, Dr. Memory suppression files, drcovlib, etc.) is to mmap the file and then parse it as a const string

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+    }
+    buf[size_read - 1] = '\0';
+
+    std::vector<char *> *lines_list = new std::vector<char *> ();
+    lines_count = split(buf, "\r\n", lines_list); /* split buffer by lines */
+
+    global_free(buf, size_to_read, HEAPSTAT_MISC);
+
+    if (lines_count <= 0) {
+        VNOTIFY(1, "An empty config file was specified, config is not used" NL);
```
derekbruening: As above, is this visible in release build?  It should be.  The name "VNOTIFY" sounds like a debug-only thing but I'm not sure what it is here.

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+
+    std::vector<char *>::iterator it;
+    for (it = lines_list->begin(); it != lines_list->end(); it++) {
+        if (*it == NULL)
+            continue;
+        /* XXX: we have to describe a format of the config file in the drltrace's
+         * documentation as well as list supported types.
+         */
+        if (parse_line(*it, line_num) == false) {
+            VNOTIFY(1, "incorrect format for the line %s in config file" NL, *it);
```
derekbruening: Should be visible

drltrace/drltrace.cpp
derekbruening: It is simple enough to add a try-except, or add a XXX comment

drltrace/drltrace.config
```diff
@@ -0,0 +1,8 @@
+; The syntax of this config file is simple. To add new function, it's enough to specify
```
mxmssh: Done

drltrace/drltrace.config
```diff
@@ -0,0 +1,8 @@
...
+; each argument of the function, function name itself and return type separated by commas.
```
mxmssh: Done

mxmssh: Well, there are several other examples which are not possible to describe using this simple language. I wanted to keep this commit small. Of course I'm planning to improve our syntax soon. Added a reference to i#1948 where we have a bit more details about this.

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+ */
+
+#define LIBCALLS_TABLE_HASH_BITS 6
+/* We init the following hashtable to be able to get library call arguments when
+ * it is required for printing.
+ */
+static hashtable_t libcalls_table;
+
+static void
+free_args_list(void *p) {
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+    std::vector<drsys_arg_t *> *args_list = (std::vector<drsys_arg_t *> *) p;
+    std::vector<drsys_arg_t *>::iterator it;
+    for (it = args_list->begin(); it != args_list->end(); it++)
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+
+static void
+init_libcalls_hashtable() {
+    hashtable_init_ex(&libcalls_table, LIBCALLS_TABLE_HASH_BITS, HASH_STRING_NOCASE,
+                      false/*!strdup*/, false, free_args_list, NULL, NULL);
+}
+
+static void
+libcalls_hashtable_delete() {
+    return hashtable_delete(&libcalls_table);
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+}
+
+static std::vector<drsys_arg_t *> *
+libcalls_search(const char *name) {
+    return (std::vector<drsys_arg_t *> *) hashtable_lookup(&libcalls_table, (void *)name);
+}
+
+static bool
+libcalls_hashtable_insert(const char *name, std::vector<drsys_arg_t *> *args_list) {
+    return hashtable_add(&libcalls_table, (void *)name, (void *) args_list);
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+}
+
+/****************************************************************************
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+ * Config file parsing routines
+ */
+
+static drsys_arg_t *
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+config_parse_type(std::string type_name, uint index) {
+
+    ASSERT(type_name.empty() == false, "an empty type name was provided");
```
mxmssh: Done here and below

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+
+    drsys_arg_t *arg = (drsys_arg_t *)global_alloc(sizeof(drsys_arg_t), HEAPSTAT_MISC);
+
+    /* sanitize input type*/
+    std::transform(type_name.begin(), type_name.end(), type_name.begin(), ::tolower);
```
mxmssh: Done here and updated .dox

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+    } else if (type_name.compare("byte") == 0) {
+        arg->type_name = "BYTE";
+        arg->size = sizeof(BYTE);
+    } else if (type_name.compare("bool") == 0) {
+        arg->type_name = "BOOL";
+        arg->size = sizeof(BOOL);
+    }
+#endif
+    else if (type_name.compare("char*") == 0) {
+        arg->type_name = "char ";
```
mxmssh: I just wanted to have a space in "char *" string.

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+        arg->type = DRSYS_TYPE_CSTRING;
+    } else if (type_name.compare("wchar*") == 0) {
+        arg->type_name = "wchar_t ";
+        arg->mode = DRSYS_PARAM_IN;
+        arg->type = DRSYS_TYPE_CWSTRING;
+    } else if (type_name.find("*") != std::string::npos) {
+        arg->size = sizeof(void *);
+        arg->mode = DRSYS_PARAM_IN;
+        arg->type_name = "<unknown> ";
+        VNOTIFY(2, "found pointer to unknown type %s in the config file" NL,
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+                type_name.c_str());
+    } else {
+        arg->type_name = "<unknown>";
+        arg->mode = DRSYS_PARAM_IN;
+        VNOTIFY(2, "found unknown type %s in the config file" NL, type_name.c_str());
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+
+    return arg;
+}
+
+static int
+split(char *line, const char *delim, std::vector<char *> *tokens_list) {
+    char *token = NULL;
+    int count = 0;
+
+    token = strtok(line, delim);
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+    if (line == NULL)
+        return false;
+
+    if (strlen(line) <= 0 || line[0] == '\n' || line[0] == '\r')
+        return false;
+
+    if (line[0] == ';') /* just a comment */
+        return true;
+
+    std::vector<char*> *tokens = new std::vector<char*>();
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+        VNOTIFY(1, "unable to parse config file at line %d: %s" NL, line_num, line);
+        delete (tokens);
+        return false;
+    }
+
+    std::vector<drsys_arg_t *> *args_vector = new std::vector<drsys_arg_t *>();
+    std::vector<char*>::iterator it;
+    for (it = tokens->begin(); it != tokens->end(); it++) {
+        /* i#1948: Currently, we don't support ret value printing and
+         * skipping it here. */
```
mxmssh: Done

drltrace/drltrace.cpp
```diff
@@ -79,6 +80,276 @@ static file_t outf;
...
+        } else if (elem_index == 1)
+            func_name = *it;
+
+        elem_index++;
+    }
+
+    delete tokens;
+
+    if (func_name == NULL || args_vector->size() <= 0) {
+        VNOTIFY(1, "unable to parse config file at line %d: %s" NL, line_num, line);
```
mxmssh: Done

drltrace/drltrace.cpp
mxmssh: Done

drltrace/drltrace.cpp
mxmssh: No, it turns that all these messages are not visible in release build. I will fix it in a separate CL, added comment for now.

drltrace/drltrace.cpp
mxmssh: Done

drltrace/drltrace_options.h
```diff
@@ -64,6 +64,12 @@ static droption_t<int> op_max_args
  "Maximum number of arguments to print.  This option allows to limit the number of "
  "arguments to be printed.  Specify 0 to disable args printing (including unknown).");
 
+static droption_t<std::string> op_config_file
+(DROPTION_SCOPE_CLIENT, "config", "./drltrace.config", "A path to drltrace's config file",
```
mxmssh: Done

drltrace/drltrace_options.h
```diff
@@ -64,6 +64,12 @@ static droption_t<int> op_max_args
...
+ "Specify a path where config is located. The default value is \"./drltrace.conf\""
+ "(current dir).  Config file describes a prototype of library functions to be able"
```
mxmssh: Ok, let's use install dir, it should also fix the problem with drltrace test failure on Appveyor. Done.

mxmssh: Done

drltrace/drltrace_options.h
```diff
@@ -64,6 +64,12 @@ static droption_t<int> op_max_args
...
+ "to print libcall arguments.  See drltrace documentation for more details.");
```
mxmssh: Our documentation for drltrace was a bit out of date. I've updated it.

drltrace/drltrace_options.h
```diff
@@ -81,6 +87,10 @@ static droption_t<bool> op_version
 static droption_t<unsigned int> op_verbose
 (DROPTION_SCOPE_ALL, "verbose", 1, "Change verbosity.", "Change verbosity.");
 
+static droption_t<bool> op_disable_config
+(DROPTION_SCOPE_CLIENT, "disable_config", false, "Disable config file usage",
```
mxmssh: Done

drltrace/drltrace_options.h
```diff
@@ -81,6 +87,10 @@ static droption_t<bool> op_version
...
+ "This option is used to disable config file usage for libcall arguments printing");
```
mxmssh: Done

drltrace/drltrace.cpp
mxmssh: Done

mxmssh:
I had a problem with droptions.h when split drltrace.cpp into two files, created drltrace_options.cpp and drltrace_options.h to make all parameters visible in the client and frontend.

drltrace/drltrace.config
derekbruening: Hmm, if we know this syntax won't support some types, it seems best to consider now whether some other syntax would be better.

There should be a version number to allow changing or extending the syntax, if the idea is to start with something we know is going to need expansion (even if not, versioning is always a good idea).

It also seems best to mention that some types are not yet supported.

mxmssh: Well, the syntax covers most of the API calls quite well, we just don't support some cases yet e.g. arguments used for output and this example your provided above which is possible to handle.
I would prefer to stay with this syntax, add versioning (I will add this together with other fixes after the second round of review) and describe not yet supported functionality that we want to add.

derekbruening: If handling a function type means we completely change the syntax, and we end up changing it to new syntax which is just as simple to write and parse as the proposed syntax here (say, using a different separator from a comma), shouldn't we just start with the new syntax?

derekbruening: I'm talking here only about fundamental issues with the proposed syntax, mainly the choice of using comma as a separator.  Adding annotations, as you mention for output params, later makes perfect sense.

mxmssh: ltrace uses standard C function prototypes e.g. int func_a (int arg1, void *arg2); and it covers almost everything (you can even define your own types and enumerations), however it is more complicated to parse such definition. My first idea was just to use ltrace syntax and don't worry about complexity of the code. Then I decided to use commas. Do you think we will have problems with commas right ? because I don't see any problems at this moment.

derekbruening: As I pointed out in my first comment, a function type with more than one argument has commas inside it.  Some character that cannot appear in a type should be chosen rather than a comma.

mxmssh: Ok, now I understand, what do you think about using forbidden in C/C++ symbols: $ or @ as a delimeter ?

derekbruening: That should work, or how about a pipe | which IMHO is nicer visually

mxmssh: agree, will prepare commit

mxmssh:
still have the problem with test, will fix it after review.

drltrace/drltrace_frontend.cpp
```diff
@@ -302,11 +308,25 @@ _tmain(int argc, const TCHAR *targv[])
         tmp--;
     *(tmp+1) = '\0';
 
+    /* in case of default config option, we use drltrace's install path */
+    if (op_config_file.get_value().compare(DEFAULT_CONFIG_DIR) == 0) {
```
derekbruening: Rather than this odd sentinel value, why not just have the default be relative to the binary as Dr. Memory does for its config files (see obtain_configfile_path())?  That seems cleaner.

drltrace/drltrace_frontend.cpp
```diff
@@ -302,11 +308,25 @@ _tmain(int argc, const TCHAR *targv[])
...
+        dr_snprintf(tmp_path, BUFFER_SIZE_ELEMENTS(tmp_path), "%s/drltrace.config",
+                    full_frontend_path /*install dir*/);
```
derekbruening: We always call NULL_TERMINATE_BUFFER after snprintf, esp when the return value is not checked

drltrace/drltrace_libcalls.cpp
```diff
@@ -0,0 +1,303 @@
...
+    /* init arg */
+    arg->type = DRSYS_TYPE_UNKNOWN;
+    arg->ordinal = index;
+    arg->size = 0;
+    arg->arg_name = NULL;
+    arg->type_name = NULL;
+    arg->pre = true;
+    arg->mode = DRSYS_PARAM_INLINED; /* considering inlined param by default */
+
+    /* i#1948: Currently, we have only few cross-platform libcalls in the config
```
derekbruening: Looks like it's missing a FIXME prefix

drltrace/drltrace_libcalls.cpp
```diff
@@ -0,0 +1,303 @@
...
+     * file which is possible to use both for Windows and Linux. However, we
+     * need to separate them into two configs and fix CMAKE accordingly.
+     * Moreover, we have to provide two different interfaces for type parsing
+     * in Windows and Linux.
+     */
+
+    /* i#1948: We have to extend a list of supported types here. */
```
derekbruening: Again needs a prefix: XXX

drltrace/drltrace_libcalls.cpp
```diff
@@ -0,0 +1,303 @@
...
+static int
+split(const char *buf, const char delim, std::vector<const char *> *tokens_list)
+{
+    int count = 0;
+    std::stringstream ss;
+    std::string item;
+
+    ss.str(buf);
+    while (std::getline(ss, item, delim)) {
+        tokens_list->push_back(strdup(item.c_str()));
```
derekbruening: Doesn't seem very C++-ish to go to the C string and strdup it

drltrace/drltrace_libcalls.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    if (tokens_count <= 0) {
+        VNOTIFY(0, "unable to parse config file at line %d: %s" NL, line_num, line);
+        return false;
+    }
+
+    std::vector<drsys_arg_t *> *args_vector = new std::vector<drsys_arg_t *>();
+    std::vector<const char *>::iterator it;
+    for (it = tokens.begin(); it != tokens.end(); ++it) {
+        /* i#1948: Currently, we don't support ret value printing and
```
derekbruening: FIXME

drltrace/drltrace_libcalls.cpp
```diff
@@ -0,0 +1,303 @@
...
+        if (!parse_line(*it, line_num)) {
+            VNOTIFY(0, "incorrect format for the line %d: %s in config file" NL,
+                    line_num, *it);
+            op_use_config.set_value(false);
+            libcalls_hashtable_delete();
+            break;
+        }
+        line_num++;
+    }
+}
```
derekbruening: missing newline

drltrace/CMakeLists.txt
```diff
@@ -133,29 +137,45 @@ if (BUILD_TOOL_TESTS)
   get_target_path_for_execution(drltrace_path drltrace)
   get_target_path_for_execution(app_path drsyscall_app)
   prefix_cmd_if_necessary(drltrace_path OFF ${drltrace_path})
-
   add_test(drltrace ${drltrace_path} -- ${app_path})
   add_test(drltrace_libcalls ${drltrace_path} -logdir - -print_ret_addr -- ${app_path})
   add_test(drltrace_symargs ${drltrace_path} -logdir - -num_max_args 4 -- ${app_path})
 
+  set(path_to_config ${INSTALL_BIN}/drltrace.config)
```
mxmssh: Done and added install for config file.

drltrace/CMakeLists.txt
```diff
@@ -133,29 +137,45 @@ if (BUILD_TOOL_TESTS)
...
+
+  # copy drltrace.config required to test libcall arguments printing
```
mxmssh: I don't understand this suggestion, currently I'm launching drltrace for tests from bin where we also copy configuration file. Do we want to copy config in build dir and launch drltrace from there ?

drltrace/drltrace.config
```diff
@@ -0,0 +1,14 @@
+; The syntax of this config file is simple. To add a new function, it's enough to specify
```
mxmssh: done

drltrace/drltrace.dox
```diff
@@ -35,47 +35,70 @@
 ***************************************************************************
 \page page_drltrace Library Tracing Tool
 
-drltrace is a standalone tool for Windows based on the Dr. Memory framework that records
-calls to shared libraries. It reports all invocations of exported functions in all shared
-libraries loaded into the target process.
+drltrace is a standalone tool based on the Dr. Memory framework that records calls to
+shared libraries. It reports all invocations of exported functions in all shared libraries
+loaded into the target process. The tool also records library call arguments and allows
+to print them in a log file.
```
mxmssh: done

drltrace/drltrace.dox
```diff
@@ -35,47 +35,70 @@
...
 
-The runtime options for this tool include:
+Similar to Linux ltrace tool, drltrace supports an external configuration file where a
```
mxmssh: done

drltrace/drltrace.h
```diff
@@ -0,0 +1,56 @@
...
+ */
+#define VNOTIFY(level, ...) do {          \
+    NOTIFY_COND(op_verbose.get_value() >= level, f_global, __VA_ARGS__); \
+} while (0)
+
+#define OPTION_MAX_LENGTH MAXIMUM_PATH
+
+void parse_config(void);
+std::vector<drsys_arg_t *> *libcalls_search(const char *name);
+void libcalls_hashtable_delete();
```
mxmssh: done

drltrace/drltrace_frontend.cpp
mxmssh: It looks like my comments here and below are wrong. Actually I'm setting drltrace's binary path (not install path) here, fixed comments. I've added a new boolean option similarly to the Dr.Memory to disable/enable default config file usage.

drltrace/drltrace_frontend.cpp
mxmssh: done here and in several other places.

drltrace/drltrace_libcalls.cpp
mxmssh: done

drltrace/drltrace_libcalls.cpp
mxmssh: done, I'm wondering what is a difference between FIXME and XXX ?

drltrace/drltrace_libcalls.cpp
mxmssh: Absolutely, we don't even need C string here. Done.

drltrace/drltrace_libcalls.cpp
mxmssh: done

drltrace/drltrace_libcalls.cpp
mxmssh: done

mxmssh:
Fixes most of the comments, have a question about running drltrace test from build dir.

drltrace/CMakeLists.txt
derekbruening: My point here is that this comment implies that you only copy the config file for your test, when it's needed for regular execution and not just the test.

drltrace/drltrace_libcalls.cpp
derekbruening: It is explained in the code style docs https://github.com/DynamoRIO/dynamorio/wiki/Code-Style

drltrace/drltrace.config
```diff
@@ -1,3 +1,38 @@
+; ***************************************************************************
```
derekbruening: IMHO # be used for a comment instead of ; as # is more commonly used that way in config files and scripts.

drltrace/drltrace_options.cpp
```diff
@@ -66,12 +66,16 @@ droption_t<int> op_max_args
...
+ "Use config file that comes with drltrace and located in the same path. Specify "
+ "no_use_config and provide a path to custom config file using -config option.");
+
 droption_t<std::string> op_config_file
-(DROPTION_SCOPE_ALL, "config", DEFAULT_CONFIG_DIR, "The path to drltrace's"
- " config file.",  "Specify a path where config is located. The default value is"
- " \"<install_dir>/drltrace.conf\" (install dir).  The config file describes the "
- " prototype of library functions for printing library call arguments.  See drltrace"
- " documentation for more details.");
+(DROPTION_SCOPE_ALL, "config", "", "The path to drltrace's config file.",
```
derekbruening: "drltrace's config file" is misleading: this is a user-provided config file, not one belonging to the drltrace installation.

mxmssh:
I will fix suggestions in the next commit soon because I don't want to re-run tests.

drltrace/drltrace_options.cpp
mxmssh: Will fix in the next commit very soon.

derekbruening: ?  "In the next commit"?  This CL should be finished before committing.

derekbruening:
> I will fix suggestions in the next commit soon because I don't want to re-run tests.

This is not the right approach: the source history and commit modularity are much more important.  If we need to revert, just to understand the history, we do not want to mix little tweaks from one commit into an unrelated future commit.

mxmssh:
Ah, sorry, it makes sense, will take it into account for future. Do we want to revert the pushed commit ?

derekbruening:
No, doesn't seem worth it, at this point it seems best to submit changes to this commit as their own separate commit.  Something should be done about the Android failures as well.

PR #2007, opened 2017-07-05
i#1948 drltrace libcall args printing: android tests and config fix

mxmssh:
Changed comment delimeter from ; to # in config file
Added config file copying on Android device
Fixed description of drltrace arguments for config file


mxmssh:
Ok, I've removed fix for i#2001 from this CL. Will prepare a separate CL.

Issue #2006, opened 2017-07-02
new drltrace tests fail on Android


mxmssh:
According to CDash the following tests are failed:
drltrace_libcalls
drltrace_symargs

PR #2008, opened 2017-07-05
i#2006: fixes drltrace test failures on Android

mxmssh:
Added copying of config file on Android device.

Fixes #2001


derekbruening:
The issue number does not look correct: do you mean #2006?

mxmssh:
Sure, fixed

Issue #2019, opened 2017-08-24
kernel module test fails with ninja generator


derekbruening:
The test assumes that make is in use:

```
$ ninja
[61/155] Generating kernel_module.ko
FAILED: tests/kernel_module.ko 
cd /work/share/build-drmemory-x64-dbg/tests/kernel_module && /usr/bin/ninja-build -C /lib/modules/4.4.14-200.fc22.x86_64/build M=/work/share/build-drmemory-x64-dbg/tests/kernel_module modules
ninja: Entering directory `/lib/modules/4.4.14-200.fc22.x86_64/build'
ninja: error: loading 'build.ninja': No such file or directory
[70/155] Generating html/index.html
ninja: build stopped: subcommand failed.

$ ls /lib/modules/4.4.14-200.fc22.x86_64/build
arch   certs   drivers   fs       init  Kconfig  lib       mm              net      scripts   sound       tools  virt
block  crypto  firmware  include  ipc   kernel   Makefile  Module.symvers  samples  security  System.map  usr    vmlinux.id
```

PR #2021, opened 2017-08-26
i#2019 kernel mod test: disable for ninja

derekbruening:
Disables the kernel module test unless the generator is "Unix Makefiles" as
it assumes that "make" is in use.

Issue: #2019


derekbruening:
Filed #2022 for the flaky test failure in fuzz_buffer.

Issue #111, opened 2014-11-28
64-bit container case


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on December 10, 2010 17:57:41_

PR 408518

filing a container case to cover 64-bit porting of Dr. Memory.
the major task is re-designing the shadow memory structure and
rewriting all of the hand-coded machine instruction sequences for
accessing shadow memory.  we'll need more indirection with
some kind of page-table-style structure, and may end up needing
caches to achieve good performance: much more complex than 32-bit.

xref PR 575713

_Original issue: http://code.google.com/p/drmemory/issues/detail?id=111_

derekbruening:
Pattern support is complete and light mode is now officially supported and present in the release package.

#1782 is a blocker for shadow modes

derekbruening:
Recent commits that did not include a proper xref for auto-linking:
3c0371c i#111 x64: heap mismatch generalization (#2026)
3d8834e i#111 x64: 64-bit shadow slowpath (#2024)
e521af6 i#111 x64: stack slowpath and redzone handling (#2023)

derekbruening:
a83d8f30d02f56e09f528a607d19b8665488821b enables 64-bit full mode on Linux though we're still missing push/pop and esp fastpaths.

Issue #1504, opened 2014-11-28
syscalls_win test failed once with ERROR_ACCESS_DENIED


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on April 03, 2014 14:24:01_

This happened once in a pre-commit run of r1861 .  This is over
rdesktop on win7x64 if that matters.

drheapstat-dbg-32: all 28 tests passed
drheapstat-rel-32: all 28 tests passed
drmemory-dbg-32: 85 tests passed, ***\* 1 tests failed: ****
        syscalls_win =>   stderr failed to match ~~Dr.M~~ 1 unique, 1 total uninitialized
                  access(es), found ~~Dr.M~~ 2 unique, 2 total uninitialized access(es)
                  instead
drmemory-dbg-64: all 8 tests passed
drmemory-rel-32: 85 tests passed, ***\* 1 tests failed: ****
        syscalls_win =>   stderr failed to match ~~Dr.M~~ 1 unique, 1 total uninitialized
                  access(es), found ~~Dr.M~~ 2 unique, 2 total uninitialized access(es)
                  instead
drmemory-rel-64: all 8 tests passed
final package: build successful; no tests for this build

Error `#1`: UNINITIALIZED READ: reading 0x001dfd50-0x001dfd51 1 byte(s) within 0x001dfd50-0x001dfd54
#0 system call NtUserGetCursorInfo
    <system call>
#1 test_sysarg_size_in_field (0x00b21729 <syscalls_win.exe+0x1729>)
    d:\derek\drmemory\git\src\tests\syscalls_win.cpp:47
#2 main             (0x00b216e8 <syscalls_win.exe+0x16e8>)
    d:\derek\drmemory\git\src\tests\syscalls_win.cpp:60
Note: @0:00:04.346 in thread 5300

Error `#2`: UNINITIALIZED READ: reading 0x001dfd24-0x001dfd28 4 byte(s) within 0x001dfd24-0x001dfd28
#0 system call NtUserGetIconInfo parameter value `#0`
    <system call>
#1 USER32.dll!GetIconInfo (0x772a4a00 <USER32.dll+0x24a00>)
    ??:0
#2 test_sysarg_size_in_field (0x00b2179a <syscalls_win.exe+0x179a>)
    d:\derek\drmemory\git\src\tests\syscalls_win.cpp:53
#3 main             (0x00b216e8 <syscalls_win.exe+0x16e8>)
[...]

PR #2028, opened 2017-08-27
i#111 x64: preliminary full mode fastpath support

derekbruening:
Generalizes the full mode fastpath to handle 2-byte shadows and scratch
registers.  Defines the top 32 bits of 32-bit written registers.  Includes
numerous other miscellaneous fixes.

Some common instructions are not yet on the fastpath, including push and
pop.

Disables -share_xl8 for now for x64 (i#2025 covers getting to work).
Disables -esp_fastpath for now for x64 (i#2027 covers implementing it).

Does not yet enable tests: that is coming soon.

Issue: #111


derekbruening:
appveyor failures are perturb_FLAKY and syscalls_win.  These passed on the prior patch here and are known to be flaky: #1504, etc.

PR #2037, opened 2017-08-29
i#1741 wrap_wincrt: mark test as flaky

derekbruening:
The wrap_wincrt test is flaky and is failing often enough that we ignore
its failures for now.

Issue: #1741


derekbruening:
Filed #2041 on the rip-rel failure for x64.

PR #2042, opened 2017-08-29
i#2039: Opts out of auto-predication

### LINKED ISSUE #2039: Auto-predication in DrMemory
On my chromebook, almost all of the tests are failing, with the following error: `Usage error: auto-predication does not support cti's`

These failures are due to dynamorio/dynamorio#1723, since we updated the DynamoRIO submodule. For now we should globally disable auto predication in all insertion events in the repo.

toshipiazza:
To fix immediate test errors due to poor interaction between global
auto-predication (introduced in dynamorio/dynamorio#1723) and meta
control flow, we pre-emptively opt out of auto-predication.

Fixes #2039

### DIFF
--- a/common/alloc.c
+++ b/common/alloc.c
@@ -7086,6 +7086,7 @@ alloc_event_bb_insert(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst
     app_pc pc = instr_get_app_pc(inst);
     if (pc == NULL)
         return DR_EMIT_DEFAULT;
+    drmgr_disable_auto_predication(drcontext, bb);
 #ifdef WINDOWS
     if (instr_get_opcode(inst) == OP_int &&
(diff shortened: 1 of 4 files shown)


common/alloc.c
```diff
@@ -7086,6 +7086,7 @@ alloc_event_bb_insert(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst
     app_pc pc = instr_get_app_pc(inst);
     if (pc == NULL)
         return DR_EMIT_DEFAULT;
+    drmgr_disable_auto_predication(drcontext, bb);
```
derekbruening: Including a comment explaining why we're disabling and whether it's meant to be temporary (plus the issue # if temporary) will improve code readability.

toshipiazza: Done.

Issue #1726, opened 2015-06-02
port to ARM container case


derekbruening:
Container case for porting Dr. Memory to ARM.

derekbruening:
I put in a series of CL's to get going on the ARM port but have to de-prioritize it due to unexpected other projects so I'm summarizing the state here:

```
1d5e794 i#1726 port to ARM: port pattern code to ARM, part 1
9d92e3e i#1726 port to ARM: port instru and disable annotation + umbra
5187b1c i#1726 port to ARM: refactor instrumentation control code
91905db i#1726 port to ARM: refactor scratch reg code
65a5dd3 i#1726 port to ARM: port utils, leak, perturb, replace, shadow
090f2f7 i#1726 port to ARM: port alloc, callstack, main, syscall, and report code
7a43ace i#1726 port to ARM: implement atomic operations for ARM
feb32db i#1726 port to ARM: update DR for instr_is_exclusive_store()
d101dbc i#1726 port to ARM: port drsyscall to ARM, part 3
a660362 i#1726 port to ARM: port drsyscall to ARM, part 2
9d72b21 i#1726 port to ARM: port drsyscall to ARM, part 1
2b8dcf0 i#1726 port to ARM: port asm_utils to ARM
0c6fb26 i#1726 port to ARM: update IF_ defines
19c4335 i#1726 port to ARM: update DR to fix asm define
e530c96 i#1726 port to ARM: update DR for extensions on ARM
```

The initial target is pattern mode.  The plan is to disable readwrite.c, fastpath.c, and stack.c, at which point every file compiles except spill.c.  I was considering swapping in drreg.  Then we need to update the pattern-mode instrumentation for ARM (load into reg before compare, and handle register lists).

derekbruening:
Xref https://github.com/DynamoRIO/drmemory/wiki/ARM-Port

derekbruening:
More refactoring work:

```
45a8c61 i#1726 port to ARM: build tests for ARM
4586123 i#1726 port to ARM: rename readwrite to slowpath
31adefa i#1726 port to ARM: refactor slowpath.c into slowpath_x86.c
ef1fa41 i#1726 port to ARM: add slowpath_arm.c template
4d9d623 i#1726 port to ARM: refactor stack.c into stack_{x86,arm}.c
614f1b3 i#1726 port to ARM: refactor fastpath.c into fastpath_{x86,arm}.c
[...]

Issue #1468, opened 2014-11-28
add Umbra tests


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on March 06, 2014 16:48:04_

We do not have any targeted tests of the Umbra library.  This issue covers creating some.

_Original issue: http://code.google.com/p/drmemory/issues/detail?id=1468_

PR #2044, opened 2017-08-30
i#1726: Port Umbra to ARM

toshipiazza:
Pulls out shadow_table_insert_app_to_shadow() into 2 arch-specific
methods for x86 and for ARM. Also implements this method for ARM.

Issue: #1726


umbra/umbra_x86.c
```diff
@@ -295,11 +300,85 @@ shadow_table_insert_app_to_shadow(void *drcontext,
...
+{
+    reg_id_t reg_idx = regs[0];
+    reg_id_t reg_tmp = regs[1];
+
+    uint disp_hi = (uint)(map->shadow_table) & 0xffff0000;
+    uint disp_lo = (uint)(map->shadow_table) & 0x0000ffff;
+    disp_hi >>= 0x10;
+
+    /* %reg_idx = table */
+    PRE(ilist, where, INSTR_CREATE_movw(drcontext,
```
toshipiazza: INSTR_CREATE_load_int() does not load 32-bit ints, so I resorted to using movw and movt. Is there a shorthand for this in the repo already? If not, should there be one?

toshipiazza:
Some questions:

1. Should we split the files up even more? We're basically using all the same code between x86 and ARM, so maybe it's not worth.
2. Is the new code sequence alright? It looks non-optimal because it requires a second scratch register, but I haven't been able to get it to use a single scratch register.

I think Umbra on x86 and ARM should now be functionally equivalent, but there are no tests in the repo for it; should I add some tests?

umbra/umbra_x86.c
derekbruening: There is instrlist_insert_mov_immed_ptrsz()

derekbruening:
> I think Umbra on x86 and ARM should now be functionally equivalent, but there are no tests in the repo for it; should I add some tests?

This is a hole is the test suite: see #1468.  We would love to have some tests added.

umbra/umbra_x86.c
toshipiazza: Whoops, forgot about that method. Done.

toshipiazza:
I'll add some tests in with this CL then.

umbra/umbra_x86.c
```diff
@@ -295,11 +300,78 @@ shadow_table_insert_app_to_shadow(void *drcontext,
...
+int
+umbra_num_scratch_regs_for_translation_arch(void)
+{
+    return 2;
 }
 
+/* code sequence:
+ * %reg_idx   = table
+ * %reg_tmp   = (%reg_addr >> 0x10)
+ * %reg_idx  += (%reg_tmp  << 0x02)
```
derekbruening: You could get rid of reg_tmp by adding (%reg_addr >> 14) here and then zeroing the bottom 2 bits, right?

toshipiazza: Sorry about the delay. I think this would only work if the static table already has the bottom 2 bits cleared:

```
reg_idx  =  0x41414141
reg_idx += (reg_addr >> 14)
reg_idx &= 0xfffffffc
```

If we simply clear the last two bits of `reg_idx`, we might kill 2 bits in the address table. I guess it's ok to assume that the static table is 4 byte aligned though, so maybe this is fine.

toshipiazza: Unrelated: there's a confusing variable called `static_shadow_table_is_used` which is statically set to false, though really the variable should be named `static_shadow_table_unused` or sthg. Can update this.

umbra/umbra_x86.c
```diff
@@ -249,22 +249,27 @@ shadow_table_app_to_shadow(umbra_map_t *map, app_pc app_addr)
...
-                                  umbra_map_t *map,
-                                  instrlist_t *ilist,
-                                  instr_t *where,
-                                  reg_id_t reg_addr,
-                                  reg_id_t reg_index)
-{
-#ifdef X86
+shadow_table_insert_app_to_shadow_arch(void *drcontext, umbra_map_t *map,
+                                       instrlist_t *ilist, instr_t *where,
+                                       reg_id_t reg_addr, reg_id_t *regs)
```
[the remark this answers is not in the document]
toshipiazza: Since we can do this with 1 reg, the array is no longer passed in.

umbra/umbra_x86.c
```diff
@@ -295,11 +300,78 @@ shadow_table_insert_app_to_shadow(void *drcontext,
...
+ * %reg_addr  >>= map->scale
+ * %reg_addr   += %reg_idx
+ */
+static void
+shadow_table_insert_app_to_shadow_arch(void *drcontext, umbra_map_t *map,
+                                       instrlist_t *ilist, instr_t *where,
+                                       reg_id_t reg_addr, reg_id_t *regs)
+{
+    reg_id_t reg_idx = regs[0];
+    reg_id_t reg_tmp = regs[1];
```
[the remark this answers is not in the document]
toshipiazza: Since we can do this with 1 reg, the array is no longer passed in.

umbra/umbra_x86.c
derekbruening: Hmm yeah that seems to never be written to?

derekbruening:
I think most of the failures are the same ones that hit DR when Travis updated their VM image about 2 weeks ago: https://github.com/DynamoRIO/dynamorio/issues/2632, https://github.com/DynamoRIO/dynamorio/issues/2634, https://github.com/DynamoRIO/dynamorio/issues/2641

I fixed the first two but am still working on the last one: if anyone can reproduce it locally that would help a great deal as it is slow going trying to diagnose through just what I can print out on Travis.

Xref #2050 where @mxmssh is also hitting failures and about to update DR and I think hitting some further warnings from the clang upgrade.

umbra/umbra_x86.c
toshipiazza: Yup looks like it's only being used once to determine whether to allocate the shadow table using `nonheap_alloc()` (if `static_shadow_table_is_used` is *true*), but uses the static shadow table if `static_shadow_table_is_used` is *false*, hence the confusion.

make/git/hook-pre-commit.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
```
toshipiazza: On my box, /bin/sh -> /bin/dash, and I believe that this script uses a bash-ism. Not sure why I only saw this recently though. Should I change this back?

tests/framework/CMakeLists.txt
```diff
@@ -29,44 +29,14 @@ set(CMAKE_BUILD_TYPE "Debug")
...
-  set(arch_defs "WINDOWS")
-endif (UNIX)
-
 find_package(DrMemoryFramework PATHS ${framework_dir})
 
 # As we'll be calling configure_DynamoRIO_{client,standalone} from within
 # a function scope, we must set the global vars ahead of time:
 configure_DynamoRIO_global(OFF ON)
 
-# We save pdb+dll space by not using libc (xref DRi#714).
```
toshipiazza: Umbra pulls in libc, so we can't do this anymore.

toshipiazza:
I only added 2 tests, one of which tests `umbra_{read,write}_shadow_memory()`. I'd also like to test `umbra_insert_app_to_shadow()` (of course). How would I go about doing this? It was mentioned [here](https://groups.google.com/forum/#!topic/dynamorio-users/-G1oTgXmfkU) that "DrMem detects whether [a pointer from the page table points to] a special block and replaces it with a writable block at that time," but does DrMem do this in a fault handler, or is the only logic for this in `umbra_{read,write}_shadow_memory()`?

make/git/hook-pre-commit.sh
derekbruening: Making it bash seems fine: we did the same thing with git_review.sh

tests/framework/CMakeLists.txt
derekbruening: On Windows too?

tests/framework/umbra_app.c
```diff
@@ -0,0 +1,167 @@
...
+GLOBAL_LABEL(FUNCNAME:)
+#ifdef X86
+        /* push callee-saved registers */
+        PUSH_SEH(REG_XBX)
+        PUSH_SEH(REG_XBP)
+        PUSH_SEH(REG_XSI)
+        PUSH_SEH(REG_XDI)
+        sub      REG_XSP, FRAME_PADDING /* align */
+        mov 	 REG_XBP, REG_XSP
+        sub		 REG_XSP, 4
```
derekbruening: The operand alignment looks off here and on the line above.  The style guide says "place opcodes on column 8 and start operands on column 17" -- hard to tell from this view which one of lines 78 and 79 is correct.

tests/framework/umbra_app.c
```diff
@@ -0,0 +1,167 @@
...
+        END_PROLOG
+        jmp      test1
+
+        /* test 1 - write to memory */
+     test1:
+        mov      REG_XDI, UMBRA_TEST_1_ASM
+        mov      REG_XDI, UMBRA_TEST_1_ASM
+        mov		 [REG_XBP], REG_XDI
```
derekbruening: Alignment again

tests/framework/umbra_app.c
```diff
@@ -0,0 +1,167 @@
...
+
+        jmp      test2
+
+        /* test 2 - read from memory */
+     test2:
+        mov      REG_XDI, UMBRA_TEST_2_ASM
+        mov      REG_XDI, UMBRA_TEST_2_ASM
+        mov		 REG_XDI, [REG_XBP]
```
derekbruening: Alignment

tests/framework/umbra_app.c
```diff
@@ -0,0 +1,167 @@
...
+
+        jmp      epilog
+
+    epilog:
+        mov 	 REG_XSP, REG_XBP
```
derekbruening: Alignment

derekbruening:
> I only added 2 tests, one of which tests umbra_{read,write}_shadow_memory(). I'd also like to test umbra_insert_app_to_shadow() (of course). How would I go about doing this? It was mentioned here that "DrMem detects whether [a pointer from the page table points to] a special block and replaces it with a writable block at that time," but does DrMem do this in a fault handler, or is the only logic for this in umbra_{read,write}_shadow_memory()?

For inlined instrumentation, it is a fault handler: e.g., https://github.com/DynamoRIO/drmemory/blob/master/drmemory/fastpath.c#L350.  For C code it's sthg like this https://github.com/DynamoRIO/drmemory/blob/master/drmemory/shadow.c#L482

tests/framework/CMakeLists.txt
toshipiazza: I think so, as of dddcfc6aefaa936fe930043be9723f372716aaef right?

tests/framework/umbra_app.c
toshipiazza: Done. Looks like this was a tabs vs spaces confusion issue.

tests/framework/umbra_app.c
toshipiazza: Done.

tests/framework/umbra_app.c
toshipiazza: Done.

tests/framework/umbra_app.c
toshipiazza: Done.

tests/framework/CMakeLists.txt
derekbruening: That one may have unnecessarily changed Windows too.  If I re-instate `set(DynamoRIO_USE_LIBC OFF)` today and build umbra it builds fine and there are no unusual imports.  Missing symbols at runtime are usually a UNIX thing for ELF and !LD_BIND_NOW: on Windows the linker complains at link time.

toshipiazza: Gotcha, should I file an issue and fix in a later CL? Or should I just fix it in this CL (i.e. revert effects of dddcfc6 only for Windows)

toshipiazza:
> For inlined instrumentation, it is a fault handler: e.g., https://github.com/DynamoRIO/drmemory/blob/master/drmemory/fastpath.c#L350. For C code it's sthg like this https://github.com/DynamoRIO/drmemory/blob/master/drmemory/shadow.c#L482

Noted. Is there a reason 64-bit Umbra abstracts away the fault handler (i.e. the fault handler appears in umbra_64.c), but 32-bit Umbra does not? Is it due to poor interaction with DrMemory?

derekbruening:
> > For inlined instrumentation, it is a fault handler: e.g., https://github.com/DynamoRIO/drmemory/blob/master/drmemory/fastpath.c#L350. For C code it's sthg like this https://github.com/DynamoRIO/drmemory/blob/master/drmemory/shadow.c#L482
> 
> Noted. Is there a reason 64-bit Umbra abstracts away the fault handler (i.e. the fault handler appears in umbra_64.c), but 32-bit Umbra does not? Is it due to poor interaction with DrMemory?

I don't think there's a good reason: it may just be historical, that DrM has a fault handler there for other reasons.  Paging @zhaoqin who wrote umbra -- is there a reason?

toshipiazza:
The travis Mac build seems to be erroring out with the following message:

```
/usr/local/Homebrew/Library/Homebrew/brew.rb:12:in `<main>': Homebrew must be run under Ruby 2.3! You're running 2.0.0. (RuntimeError)
```

derekbruening:
> The travis Mac build seems to be erroring out with the following message:
>
> ```
> /usr/local/Homebrew/Library/Homebrew/brew.rb:12:in `<main>': Homebrew must be run under Ruby 2.3! You're running 2.0.0. (RuntimeError)
> ```

DR hit this last week and I was hoping Travis would fix it by now.  Searching shows this Travis issue which lists some workarounds (didn't read it all yet):
https://github.com/travis-ci/travis-ci/issues/8552

derekbruening:
Thanks for trying this here, I'm doing the same thing for DR: https://github.com/DynamoRIO/dynamorio/pull/2672

toshipiazza:
Is this ready for merge? Were the last few tests I added reviewed?

PR #2048, opened 2017-09-13
i#1948 library calls arguments printing:gdi32 and advapi32 dlls

mxmssh:
Added function prototypes from gdi32 and advapi32 dlls.
Fixed two or more pointers handling in headers_parser.py.
Added checking for function API name in prototypes provided by
headers_parser.py.


drltrace/scripts/gen_drltrace_config.py
```diff
@@ -353,10 +353,272 @@
 "LSTATUS":"LSTATUS",
 "DEP_SYSTEM_POLICY_TYPE":"DEP_SYSTEM_POLICY_TYPE",
 "CURRENCYFMTA":"CURRENCYFMTA",
-"va_list":"va_list"}
+"va_list":"va_list",
```
derekbruening: Would it make sense to omit those that just identity mappings and use the original name for missing entries?  That would make this list much shorter.  Having an automated rule to remove a leading P is a little harder b/c some type names start with P.

derekbruening:
What about testing?  I don't see any tests in this CL.

drltrace/scripts/gen_drltrace_config.py
mxmssh: Yes, it makes sense.

mxmssh:
I will add more tests soon.

 I see the strange problem in Travis build. 
/.../drmemory/drsyscall/table_linux.c:63:7: error: shifting a negative signed value is undefined [-Werror,-Wshift-negative-value]

I don't even touch C code at all in this CL. Why do we have such problem ?

derekbruening:
> I will add more tests soon.

It's generally best to include tests with a feature.  Separating them a) makes it less likely to find bugs in features before checking them in; b) separates pieces of code that should go together for bisecting or reverting; c) makes it more likely that tests will never be committed as other priorities come in once the feature is in.  Please consider including tests with future feature CL's.

>  I see the strange problem in Travis build. 
> /.../drmemory/drsyscall/table_linux.c:63:7: error: shifting a negative signed value is undefined [-Werror,-Wshift-negative-value]
> I don't even touch C code at all in this CL. Why do we have such problem ? 

Travis upgraded their VM's.  They completely broke the DR tests: https://github.com/DynamoRIO/dynamorio/issues/2634, https://github.com/DynamoRIO/dynamorio/issues/2632

drltrace/scripts/gen_drltrace_config.py
mxmssh: There is a problem with this idea, if I understood you right. I can not just remove mappings e.g. "LSTATUS":"LSTATUS because I usually manually analyze all new types found in the SDK files to indentify pointers. For example, if I remove mappings, it would be hard for me to understand which type I already saw and which one is new. Potentially, we can have a list of already parsed types, it will also make this map shorter (but it will increase search time).

mxmssh:
Yes, it is absolutely right, I will try to add new features using this principle.

drltrace/scripts/gen_drltrace_config.py
derekbruening: Yes, makes sense that this serves as a list of known types.

mxmssh:
> Travis upgraded their VM's. They completely broke the DR tests: DynamoRIO/dynamorio#2634, DynamoRIO/dynamorio#2632

Do we need to fix something in DrMemory to make it work ?

derekbruening:
> > Travis upgraded their VM's. They completely broke the DR tests: DynamoRIO/dynamorio#2634, DynamoRIO/dynamorio#2632

> Do we need to fix something in DrMemory to make it work ?

I haven't looked at DrM yet, weren't you the one seeing failures?  Whatever failures you're seeing, yes they'll need to be fixed.  Given that 32-bit DR was completely broken I would expect all the 32-bit Linux tests to fail until my DR fix from yesterday is imported (go ahead and update DR in DrM).

Issue #2038, opened 2017-08-29
malloc test nondet failure on Travis Mac


derekbruening:
Happened just once:
https://travis-ci.org/DynamoRIO/drmemory/builds/269437543?utm_source=email&utm_medium=notification

```
drmemory-dbg-32: 12 tests passed, **** 1 tests failed: ****
	malloc =>   stderr failed to match "~~Dr\.M~~ 1 unique, 1 total, 59 byte\(s\) of
		  leak\(s\)", found "~~Dr.M~~ 0 unique, 0 total, 0 byte(s) of leak(s)"
		  instead
```

derekbruening:
This just started happening with an update to the embedded DR:

https://travis-ci.org/DynamoRIO/drmemory/jobs/301205197
```
The following tests FAILED:
	  3 - malloc (Timeout)
```

But, that test succeeds on a local Mac machine.  I'm going to monitor it and see what happens over the next few runs.

PR #2066, opened 2017-11-29
DRi#241: use new kernel xfer events

derekbruening:
Adds use of DR's new kernel xfer events, eliminating the need to wrap the
Ki routines and watch several system calls.  Removes those events from the
alloc.h interface in favor of local kernel xfer events in the user.
Removes app signal handler tracking which is no longer needed.

Updates DR to fa465385.  Relaxes an assert triggered by DR's ea84da0
regarding handling displaced vsyscall code

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -807,7 +807,7 @@ else ()
 endif ()
 
 # when updating this, also update the git submodule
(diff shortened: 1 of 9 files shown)


derekbruening:
Travis failure is Mac malloc test again: #2038

PR #2090, opened 2018-03-11
Run Mac with -VV to diagnose why the package version number is wrong

derekbruening:


derekbruening:
OK the answer is that we deliberately change the version to satisfy clang:
```
  # clang linker disallows any but major # being >= 256 (1024 for x64) so we do
  # mod 200 (we assume we'll never confuse versions 200 apart) and add 56 (to
  # distinguish from real releases).
  math(EXPR VERSION_NUMBER_PATCHLEVEL "(${VERSION_NUMBER_PATCHLEVEL} % 200) + 56")
```

PR #2131, opened 2018-10-26
i#2128: Fix for umbra overflow in special block creation.

### LINKED ISSUE #2128: Umbra special block protection is shifted and overflows the block
https://github.com/DynamoRIO/drmemory/blob/6699e1bf92c7246770911a3affa7d6977049f985/umbra/umbra_32.c#L165

Perhaps I am reading the code incorrectly and my apologies if this is the case.

From my understanding, the setting of the _start_ param for the dr_memory_protect() call is incorrect. _block_ is incremented in _shadow_table_init_redzone()_ to refer to the shadow block skipping the upper redzone. However, the set size of the function call is that in **map->shadow_block_alloc_size**, which also includes redzone sizes. Consequently, the upper redzone of the shadow block is not being protected. Does anyone agree?

I will try and program a quick check to test my suspicion.

johnfxgalea:
Fixes overflow when setting protection for special block.

Fixes #2128

### DIFF
--- a/umbra/umbra_32.c
+++ b/umbra/umbra_32.c
@@ -162,7 +162,8 @@ shadow_table_create_special_block_helper(umbra_map_t     *map,
 
     /* we will never write to the special */
     IF_DEBUG(ok = )
-        dr_memory_protect(block, map->shadow_block_alloc_size, DR_MEMPROT_READ);
+        dr_memory_protect(block - map->options.redzone_size,
+                          map->shadow_block_alloc_size, DR_MEMPROT_READ);
     ASSERT(ok, "-w failed: will have inconsistencies in shadow data");
     LOG(UMBRA_VERBOSE, "created new shadow special block "PFX"\n", block);
     special_block->start = block;


derekbruening:
2 comments on the commit message:

nit: missing colon after the issue number in the title

If this fixes the issue (it does, right?), use the syntax `Fixes #2128` on a line by itself.  (And the syntax for a non-complete fix is `Issue: #2128`.)

johnfxgalea:
> 2 comments on the commit message:
> 
> nit: missing colon after the issue number in the title
> 
> If this fixes the issue (it does, right?), use the syntax `Fixes #2128` on a line by itself. (And the syntax for a non-complete fix is `Issue: #2128`.)

Do I add Fixes #2128 also in the code, as a comment?

derekbruening:
On Fri, Oct 26, 2018 at 6:16 PM John F.X. Galea <notifications@github.com>
wrote:

> 2 comments on the commit message:
>
> nit: missing colon after the issue number in the title
>
> If this fixes the issue (it does, right?), use the syntax Fixes #2128 on
> a line by itself. (And the syntax for a non-complete fix is Issue: #2128.)
>
> Do I add Fixes #2128 <https://github.com/DynamoRIO/drmemory/issues/2128>
> also in the code, as a comment?
>

No, just in the commit message body.

Issue #2115, opened 2018-08-28
CRASH x64 free.shadow on appveyor


derekbruening:
This is the master run after I put in -no_vm_base_near_app, though the PR passed:

https://ci.appveyor.com/project/DynamoRIO/drmemory/build/1.0.247
```
15/56 Test #15: free.shadow .......................***Failed    2.00 sec
247STDOUT: 
248STDERR: <Starting application C:\projects\drmemory\build\build_drmemory-dbg-64\tests\free.exe (2444)>
249<Initial options = -no_dynamic_options -logdir 'C:\projects\drmemory\build\build_drmemory-dbg-64\logs\dynamorio' -client_lib 'C:\projects\drmemory\build\build_drmemory-dbg-64\bin64\debug\drmemorylib.dll;0;`-callstack_style` `0x27` -no_results_to_stderr `-pattern` `0` -logdir `C:\projects\drmemory\build\build_drmemory-dbg-64\logs` -symcache_dir `C:\projects\drmemory\build\build_drmemory-dbg-64\logs\symcache` -lib_blacklist `C:\windows*.d??,C:\Program Files\Common Files\Microsoft Shared*.d??,C:\Program Files (x86)\Common Files\Microsoft Shared*.d??` -resfile 2444 ' -code_api -probe_api -msgbox_mask 0 -dumpcore_mask 0x87fd -stack_size 56K -disable_traces -no_enable_traces -max_elide_jmp 0 -max_elide_call 0 -no_shared_traces -bb_ibl_targets -bb_single_restore_prefix -no_shared_trace_ibl_routine -no_enable_reset -no_reset_at_switch_to_os_at_vmm_limit -reset_at_vmm_percent_free_limit 0 -no_reset_at_vmm_full -reset_at_commit_free_limit 0K -reset_every_nth_pending 0 -no_vm_base_near_app -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct -pad_jmps_mark_no_trace >
250~~Dr.M~~ Dr. Memory version 1.11.17771
251~~Dr.M~~ WARNING: 64-bit non-pattern modes are experimental on Windows
252~~Dr.M~~ Running "C:/projects/drmemory/build/build_drmemory-dbg-64/tests/free.exe"
[...]

PR #2135, opened 2018-10-31
Update non-Windows base version to 2.0

derekbruening:
The version split when we added 64-bit full mode to UNIX and released
it as 2.0.x before it was ready for Windows.  The cronbuilds were
using the 1.11.x Windows number so we could share the same Github
version, but that just caused (justifiable) confusion.  Here we bump
the version to 2.0 for UNIX but leave it at 1.11 for Windows, which
will result in different git tags and cronbuilds until we get Windows
up to 2.0.


derekbruening:
Appveyor failure is #2115: x64 free.shadow

Issue #2049, opened 2017-09-16
Unable to build DrMemory with new DR updates


mxmssh:
/home/osboxes/cmake/cmake-2.8.12-rc2-Linux-i386/bin/cmake -DCMAKE_BUILD_TYPE=Debug ../drmemory/

CMake Error at tests/CMakeLists.txt:530 (add_dependencies):
  Cannot add target-level dependencies to non-existent target
  "syscalls_unix".

  The add_dependencies works for top-level logical targets created by the
  add_executable, add_library, or add_custom_target commands.  If you want to
  add file-level dependencies see the DEPENDS option of the add_custom_target
  and add_custom_command commands.

derekbruening:
I updated DR and neither I nor Travis saw this issue

derekbruening:
Is the dependence line just missing `if (NOT X64)`?

PR #2138, opened 2018-11-09
DRi#1557: DR CMake functions now take extra arg

hgreving2304:
Adding location suffix parameter to calls of DynamoRIO_get_target_path_for_execution, DynamoRIO_copy_target_to_device. This patch does not address the LOCATION part of i#1557 itself.


CMakeLists.txt
```diff
@@ -1050,7 +1050,7 @@ endif ()
```
derekbruening: I don't see the dynamorio submodule being updated?  So this is passing extra args to functions that don't take them?  I guess CMake doesn't complain and things pass but that's not what we want.  We want to update the submodule.

derekbruening: > i#1557: compatibility update, CMake API dynamorio has changed.

The issue namespaces are different: 1557 is a different issue here.  See prior issues: they use `DRi#1557` to refer to a DR issue in a title.  DynamoRIO/dynamorio/issues/1557 will auto-link in the body.

derekbruening: > i#1557: compatibility update, CMake API dynamorio has changed.

grammar nit: comma isn't right there: should be a colon or sthg.  suggest sthg like: `DRi#1557: DR CMake functions now take extra arg` which gives an idea of what changed.

hgreving2304: Done

hgreving2304: I think I just did, please double-check if that was right (just git added dynamorio to latest)

tests/CMakeLists.txt
```diff
@@ -500,7 +500,7 @@ newtest_ex(state state.c "" "" "" OFF "" "ANY")
 if (UNIX)
   newtest(signal signal.c)
   if (NOT X64) # FIXME i#111: failing on Travis
-    newtest(syscalls_unix syscalls_unix.c)
+    #newtest(syscalls_unix syscalls_unix.c)
```
derekbruening: Please examine the diff before pushing

tests/CMakeLists.txt
```diff
@@ -536,7 +536,7 @@ if (UNIX)
           VERBATIM)
       add_custom_target("${kmod_name}"
           DEPENDS "${kmod_bin}")
-      add_dependencies(syscalls_unix "${kmod_name}")
+      #add_dependencies(syscalls_unix "${kmod_name}")
```
derekbruening: Another mistakenly included change.  Is this that #2049?  Is it just missing `if (NOT X64)?`

tests/CMakeLists.txt
hgreving2304: Done

tests/CMakeLists.txt
hgreving2304: Done

hgreving2304: Yes it might, I had changed this locally to be able to build.

Issue #2088, opened 2018-03-06
CRASH memalign.nodelay test


derekbruening:
This happened once on Travis on Linux:

https://travis-ci.org/DynamoRIO/drmemory/jobs/350004107
```
45/95 Test #45: memalign.nodelay ..................***Failed    3.51 sec
STDOUT: 
STDERR: <Starting application /home/travis/build/DynamoRIO/drmemory/build_drmemory-dbg-64/tests/memalign (19634)>
<Initial options = -no_dynamic_options -logdir '/home/travis/build/DynamoRIO/drmemory/build_drmemory-dbg-64/logs/dynamorio' -client_lib '/home/travis/build/DynamoRIO/drmemory/build_drmemory-dbg-64/bin64/debug/libdrmemorylib.so;0;`-batch` `-callstack_style` `0x27` -no_results_to_stderr `-delay_frees` `0` -logdir `/home/travis/build/DynamoRIO/drmemory/build_drmemory-dbg-64/logs` -symcache_dir `/home/travis/build/DynamoRIO/drmemory/build_drmemory-dbg-64/logs/symcache` -resfile 19634 ' -code_api -stack_size 56K -signal_stack_size 32K -disable_traces -no_enable_traces -max_elide_jmp 0 -max_elide_call 0 -no_shared_traces -bb_ibl_targets -bb_single_restore_prefix -no_shared_trace_ibl_routine -no_enable_reset -no_reset_at_switch_to_os_at_vmm_limit -reset_at_vmm_percent_free_limit 0 -no_reset_at_vmm_full -reset_at_commit_free_limit 0K -reset_every_nth_pending 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<Paste into GDB to debug DynamoRIO clients:
set confirm off
add-symbol-file '/home/travis/build/DynamoRIO/drmemory/build_drmemory-dbg-64/bin64/debug/libdrmemorylib.so' 0x000000007381d000
add-symbol-file '/home/travis/build/DynamoRIO/drmemory/build_drmemory-dbg-64/dynamorio/lib64/debug/libdynamorio.so' 0x00007f36f7a524e8
add-symbol-file '/lib/x86_64-linux-gnu/libc.so.6' 0x00007f36f7452520
add-symbol-file '/lib64/ld-linux-x86-64.so.2' 0x00007f36f720db00
>
[...]

PR #2155, opened 2019-02-24
i#1848: auto-generate a syscall file on a new OS

### LINKED ISSUE #1848: add end-user support for updating syscall #'s from pdb's
The goal is to future-proof Dr. Memory: make it more adaptive to avoid requiring manual updates to fix breakages on each new Windows change.  Xref #1826.

The plan is:
- Detect unknown version by looking at particular syscall #'s (as we can't rely on PEB versions anymore): xref https://github.com/DynamoRIO/dynamorio/issues/1598
- Create utility that downloads pdb's for the core dll's, does sthg like what winsysnums does, and comes up with new syscall numbers. We should be able to automate everything except for the usercall stuff.
- Can we launch the helper process from our online client?  Even if so, we'll need to cache the results, so we could ask the user to run the utility standalone?
- Cache the results and load them in.

Things can still break if the syscall wrappers change (xref https://github.com/DynamoRIO/dynamorio/issues/1854) or other things besides numbers change, but this would be an improvement and could help future-proof Dr. Memory.

derekbruening:
Moves the default syscall file location into logs/symbols instead of the
bin/ directory, as we need a writable location.  Adds a new option
-syscall_number_path to allow specifying a custom location.

Adds functionality to drsyscall to fetch debug info for system libraries
and parse every symbol in each library looking for syscall wrappers.  Two
interfaces are added: drsys_find_sysnum_dlls() and
drsys_generate_sysnum_file().  Includes usercall identification support via
a list of target wrappers for key usercalls plus interpolation between
known numbers to include speculative results for all usercalls.  The
syscall wrapper parsing code is based on DR's winsysnums.c, expanded to
support 64-bit.

The resulting syscall file only supports the current machine, unlike the
general files we have posted for downloading.

Adds auto-triggering of the new functionality when an unknown OS version is
detected.  A special exit code is used, and when the frontend sees it, it
invokes the drsyscall functions for generating a syscall file.  If that is
successful, it re-launches the app.

Adds new options -vv and -vvv to support raising verbosity for the
drsyscall code invoked directly from the frontend.

Changes the existing mksystable.pl-based test to test the new auto-gen
functionality.

Fixes #1848

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1376,7 +1376,8 @@ if (USE_DRSYMS)
   set(DynamoRIO_RPATH ON)
   configure_DynamoRIO_standalone(${toolname})
   set(DynamoRIO_RPATH ${old_rpath})
-  target_link_libraries(${toolname} drinjectlib drconfiglib drfrontendlib)
+  target_link_libraries(${toolname} drinjectlib drconfiglib drfrontendlib
+    drsyscall_static)
   if (WIN32)
     set_target_properties(${toolname} PROPERTIES
       VERSION ${TOOL_VERSION_NUMBER}
--- a/common/utils.c
+++ b/common/utils.c
@@ -45,7 +45,7 @@
 /* globals that affect NOTIFY* and *LOG* macros */
 int tls_idx_util = -1;
 bool op_print_stderr = true;
-uint op_verbose_level;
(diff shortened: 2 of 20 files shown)


derekbruening:
Looks like a non-det failure in the memalign test: #2088.

Issue #677, opened 2014-11-28
drmemory -leaks_only does not work with -no_esp_fastpath


derekbruening:
_From [zhao...@google.com](https://code.google.com/u/106321947286816917100/) on November 08, 2011 16:38:49_

$ ~/Workspace/DrMemory/exports/bin/drmemory.pl -no_esp_fastpath -dr_debug -debug -pause_at_assert -leaks_only  -dr ~/Workspace/DynamoRIO/exports -- tests/hello
...
<Application hello (6034).  Internal Error Internal DynamoRIO Error: /home/zhaoqin/Workspace/DynamoRIO/src/core/linux/signal.c:3833 pc != 0
(Error occurred @6 frags)
version 3.0.1133, custom build
-client_lib '/home/zhaoqin/Workspace/DrMemory/exports/drmemory/bin/bin32/debug/libdrmemory.so;0;-logdir `/home/zhaoqin/Workspace/DrMemory/exports/drmemory/logs`   -no_esp_fastpath -pause_at_
0x19dc1de8 0xf74c42df
0x19dc1f48 0xf76419fd
0x19dc1fe8 0xf76055c6
0xffb069b8 0xf771d7e9
0xffb06a08 0xf770f88f>

when using -leaks_only is used, the option options.shadowing seems turned off, so in instrument_init, no shared code is generated.

_Original issue: http://code.google.com/p/drmemory/issues/detail?id=677_

PR #2164, opened 2019-03-10
i#111 x64: support -leaks_only -no_esp_fastpath

derekbruening:
For x64 we do not yet have -esp_fastpath, but we want to support
-leaks_only.  It uses part of the -esp_fastpath path to run a stack
zeroing loop, but it does not actually need the -esp_fastpath gencode,
so we simply remove the option check and route to the fastpath
handler.

Issue: #111, #677


derekbruening:
Looks like the test failures are actually from fefc4645d3227a59093f5a9ad10beb8ac397bd9a and some install step failure wasn't caught by the wrapper and so it marked the build as green.  Going to commit this and separately try to fix the tests failures.

PR #2174, opened 2019-03-16
DO NOT CHECK IN: temporary diagnostics to try to debug the crash seen…

derekbruening:
… only on Appveyor


derekbruening:
The crash is filed as #2178

derekbruening:
Abandoning: no longer using Appveyor.  If we can repro somewhere else we can revive the analysis.

Issue #2170, opened 2019-03-12
true uninitialized read in ntdll!RtlRestoreContext


derekbruening:
On Win10 1809 I'm seeing these types of errors reported on the `state`, `cs2butMTd`, and maybe more tests in 64-bit:

```
Error #8: UNINITIALIZED READ: reading 0x000000a1860fd560-0x000000a1860fd5c0 96 byte(s) within 0x000000a1860fd4c0-0x000000a1860fd6c0
# 0 ntdll.dll!RcConsolidateFrames (0x00007ffda3783c73 <ntdll.dll+0xa3c73>) modid:6
# 1 ntdll.dll!RtlUnwindEx (0x00007ffda36e6182 <ntdll.dll+0x6182>) modid:6
# 2 _UnwindNestedFrames [f:\dd\vctools\crt\crtw32\eh\amd64\trnsctrl.cpp:482] (0x00007ff643d940ab <cs2bugMTd.exe+0x140ab>) modid:3
# 3 main          [d:\derek\drmemory\git\src\tests\cs2bug.cpp:245] (0x00007ff643d820d0 <cs2bugMTd.exe+0x20d0>) modid:3
Note: @0:00:08.687 in thread 1448
Note: instruction: fxrstor 0x00000100(%rcx)
```

state has 2 uninits but they seem related:
```
Error #2: UNINITIALIZED READ: reading 0x000000c1a54fea00-0x000000c1a54fea60 96 byte(s) within 0x000000c1a54fe960-0x000000c1a54feb60
# 0 ntdll.dll!RtlRestoreContext (0x00007ffda3783a76 <ntdll.dll+0xa3a76>)
    ??:0
# 1 ntdll.dll!RtlUnwindEx (0x00007ffda36e6182 <ntdll.dll+0x6182>)
    ??:0
# 2 ntdll.dll!RtlUserThreadStart (0x00007ffda374a251 <ntdll.dll+0x6a251>)
    ??:0
Note: @0:00:02.484 in thread 10356
Note: instruction: fxrstor 0x00000100(%rcx)

Error #3: UNINITIALIZED READ: reading 0x000000c1a54fe7da-0x000000c1a54fe7e0 6 byte(s) within 0x000000c1a54fe7d0-0x000000c1a54fe7f8
# 0 ntdll.dll!RtlRestoreContext (0x00007ffda3783b8f <ntdll.dll+0xa3b8f>)
    ??:0
[...]

Issue #2178, opened 2019-03-17
CRASH at 0x..96c1 on appveyor in various tests


derekbruening:
I can't reproduce these crashes locally on a win8.1 or other machine.

Appveyor 64-bit has crashed in registers.exe, malloc.exe, and pcache.exe.  The crashes always have the address 0x..96c1.  Here is an example:

```
[00:06:17]   8/114 Test   #7: registers .........................***Failed    1.83 sec
[00:06:17] STDOUT: 
[00:06:17] STDERR: <Starting application C:\projects\drmemory\build\build_drmemory-dbg-64\tests\registers.exe (864)>
[00:06:17] <Initial options = -no_dynamic_options -logdir 'C:\projects\drmemory\build\build_drmemory-dbg-64\logs\dynamorio' -client_lib 'C:\projects\drmemory\build\build_drmemory-dbg-64\bin64\debug\drmemorylib.dll;0;`-callstack_style` `0x27` -no_results_to_stderr -logdir `C:\projects\drmemory\build\build_drmemory-dbg-64\logs` -symcache_dir `C:\projects\drmemory\build\build_drmemory-dbg-64\logs\symcache` -lib_blacklist `C:\windows*.d??,C:\Program Files\Common Files\Microsoft Shared*.d??,C:\Program Files (x86)\Common Files\Microsoft Shared*.d??` -resfile 864 ' -code_api -probe_api -msgbox_mask 0 -dumpcore_mask 0x87fd -stack_size 56K -disable_traces -no_enable_traces -max_elide_jmp 0 -max_elide_call 0 -no_shared_traces -bb_ibl_targets -bb_single_restore_prefix -no_shared_trace_ibl_routine -no_enable_reset -no_reset_at_switch_to_os_at_vmm_limit -reset_at_vmm_percent_free_limit 0 -no_reset_at_vmm_full -reset_at_commit_free_limit 0K -reset_every_nth_pending 0 -no_vm_base_near_app -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct -pad_jmps_mark_no_trace >
[00:06:17] ~~Dr.M~~ Dr. Memory version 1.11.17967
[00:06:17] ~~Dr.M~~ Running "C:/projects/drmemory/build/build_drmemory-dbg-64/tests/registers.exe"
[...]

PR #2179, opened 2019-03-17
i#111 x64: Fix or suppress failing win64 tests

derekbruening:
Fixes tests/operators taking forever on its OOM test by tweaking the
code to avoid compiler warnings on very large allocations.

Fixes x64 bugs in handle stack operations:
+ Handle DR rip-rel mangling when decoding from the cache on
  the esp adjust slowpath.
+ Add an x64 chkstk pattern match.

Expands the i#2170 suppression to match ntdll.dll!RcConsolidateFrames.

Issue: #111, #2170


derekbruening:
Failures are #2178.

Issue #2199, opened 2019-06-22
Shadow initialisations reuse static table.


johnfxgalea:
Umbra's static table is reused on multiple instances due to an unset flag, i.e., static_shadow_table_unused.

johnfxgalea:
It must also be said that this might be an intended functionality - that multiple initialisations should use the same table. If this is the case, at the very least we need to add a line and document this behaviour.

derekbruening:
The code does look suspect, with that variable never been set, and being renamed in 
3c0c21e0 from static_shadow_table_is_used to static_shadow_table_unused -- strange.  From talking to @zhaoqin offline: the plan was to support multiple mappings and hence dynamic tables are needed but the code never made it to testing for such cases.

PR #2200, opened 2019-06-22
Fixes multi init Umbra tables

johnfxgalea:
Sets the `static_shadow_table_unused` flag to true to avoid using the same static instance upon multiple shadow table inits.

Fixes: #2199

### DIFF
--- a/umbra/umbra_32.c
+++ b/umbra/umbra_32.c
@@ -476,6 +476,7 @@ shadow_table_init(umbra_map_t *map)
                                           HEAPSTAT_SHADOW);
     } else {
         map->shadow_table = static_shadow_table;
+        static_shadow_table_unused = true;
     }
     /* sets the whole address space to default special block first */
     if (!TEST(UMBRA_MAP_CREATE_SHADOW_ON_TOUCH, map->options.flags)) {
--- a/umbra/umbra_32.c
+++ b/umbra/umbra_32.c
@@ -67,9 +67,6 @@
 #define SHADOW_TABLE_INDEX(addr)   ((ptr_uint_t)(addr) >> APP_BLOCK_BITS)
 #define SHADOW_TABLE_APP_BASE(idx) ((ptr_uint_t)(idx)  << APP_BLOCK_BITS)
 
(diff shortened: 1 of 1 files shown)


johnfxgalea:
Note, we don't need to do anything for exits due to the following code in `shadow_table_exit`:

```
if (map->shadow_table != static_shadow_table)
    nonheap_free(map->shadow_table, SHADOW_TABLE_SIZE, HEAPSTAT_SHADOW);
```

umbra/umbra_32.c
```diff
@@ -476,6 +476,7 @@ shadow_table_init(umbra_map_t *map)
                                           HEAPSTAT_SHADOW);
     } else {
         map->shadow_table = static_shadow_table;
+        static_shadow_table_unused = true;
```
derekbruening: So the idea is to use the static table for the first mapping, and if someone else wants a new mapping, to dynamically allocate.  Could we rename `s/static_shadow_table_unused/static_shadow_table_claimed/` or sthg to make this clearer and hopefully avoid things like https://github.com/DynamoRIO/drmemory/pull/2044#discussion_r140582790 where it is mis-interpreted.

To be safe we may want a mutex here as well, or else to document that this can only be called at process init with a single thread (i.e., document it's not thread-safe).

Alternatively, zhaoqin@ suggests removing the static table and the flag altogether and always using dynamic alloc, which seems fine to me as well.

johnfxgalea: I'm personally not a big fan of global variables and use them only when necessary. I like @zhaoqin's suggestion. It makes the code cleaner with less branching.

derekbruening:
pcache-use failure: have not seen that on DrM's Travis before but have on some DR nondet tests.  Filed as https://github.com/DynamoRIO/drmemory/issues/2202

johnfxgalea:
@derekbruening Seems like we hit the failure again. Shall we halt merging until further investigation?

derekbruening:
Hmm, appveyor has slowesp and noreplace_realloc failing.

derekbruening:
The tests are in poor shape.  I don't think this is involved.  Going to merge and sort out afterward.

PR #2211, opened 2019-07-06
i#1983: Add syscall auto-generation to drstrace and drltrace

### LINKED ISSUE #1983: Update drstrace and drltrace to use auto-generated syscall number files
I have a fresh installation of Dr. Memory on a Windows 10 VM. When I run
bin\drstrace -- calc
and
bin64\drstrace -- calc
in cmd with administrator privilege I receive the error
ASSERT FAILURE: D:\drmemory_package\drstrace\drstrace.c:670: false (drsys failed to init)
and no output is produced.

derekbruening:
Adds support for reading system call numbers from a text file to both
drstrace and drltrace.  Adds support for auto-generating that file,
just like is done in Dr. Memory.

Updates DR to 1713c2d2 to fix a decoder assert that breaks the
auto-generation.

Fixes #1983

### DIFF
--- a/drltrace/CMakeLists.txt
+++ b/drltrace/CMakeLists.txt
@@ -1,5 +1,5 @@
 # **********************************************************
-# Copyright (c) 2010-2017 Google, Inc.    All rights reserved.
+# Copyright (c) 2010-2019 Google, Inc.    All rights reserved.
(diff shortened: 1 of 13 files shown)


derekbruening:
Failures are #2178

PR #2224, opened 2019-09-16
i#2083: Reference comparison tables above 2GB

### LINKED ISSUE #2083: ASSERT "truncating value to (int)disp" fastpath_x86.c:3858
This shows up in the 64-bit fastpath on Windows:
```
ASSERT FAILURE (thread 6940): D:\derek\drmemory\git\src\drmemory\fastpath_x86.c:3858: CHECK_TRUNCATE_RANGE_int((ptr_int_t)shadow_dword_is_addr_not_bit) (truncating value to (int)disp)
```

And I now see it on a Linux Ubuntu-variant machine:
```
$ bin64/drmemory -- tests/free
<...>
~~Dr.M~~ Dr. Memory version 1.11.17499
~~Dr.M~~ ASSERT FAILURE (thread 156900): /home/bruening/drmemory/git/src/drmemory/fastpath_x86.c:3788: CHECK_TRUNCATE_RANGE_int((ptr_int_t)shadow_byte_addr_not_bit) (truncating value to (int)disp)
```

derekbruening:
Adds support for loading from shadow value comparison tables that cannot be
reached by a single 32-bit displacement.  To avoid needing another scratch
register, we instead use the top bits of the register holding the shadow
value, inserting the top 32 bits of the table address using a sequence
ror;or;ror, and then clearing after the comparison.

Fixes #2083

### DIFF
--- a/drmemory/fastpath_x86.c
+++ b/drmemory/fastpath_x86.c
@@ -1407,9 +1407,10 @@ insert_table_access_pre(void *drcontext, instrlist_t *bb, instr_t *inst,
         __FUNCTION__, table_addr);
     PRE(bb, inst,
         INSTR_CREATE_ror(drcontext, opnd_create_reg(base_reg), OPND_CREATE_INT8(32)));
+    int top_bits = (int)((table_addr - disp) >> 32);
     PRE(bb, inst,
         INSTR_CREATE_or(drcontext, opnd_create_reg(base_reg),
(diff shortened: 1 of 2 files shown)


derekbruening:
I believe the Appveyor failures are #2178 though the addresses changed: so maybe related to these table addresses being truncated?!?

Issue #1438, opened 2014-11-28
port to Mac OSX


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on February 09, 2014 16:21:07_

This is a container case for porting Dr. Memory to Mac OSX.

Xref porting DR to Mac: https://code.google.com/p/dynamorio/issues/detail?id=58 Some things that need to be done:
- Refactor LINUX to UNIX
- Add MACOS support to DrSyscall (xref DRr2502)

_Original issue: http://code.google.com/p/drmemory/issues/detail?id=1438_

Issue #2184, opened 2019-03-21
failure to init Umbra due to segment-spanning 2TB reservation


StephenHillAtViaivi:
Using DrMemory-Windows-2.1.0-1.msi, I see this with all my Cpputest exe's built with MSVS2017:

```
C:\Users\SHill\SHill_view\tm_build_system\build\unittest64>"C:\Program` Files (x86)\Dr. Memory\bin64\drmemory.exe" -batch -v -- seg_unittest.exe
INFO: targeting application: "C:\Users\SHill\SHill_view\tm_build_system\build\unittest64\seg_unittest.exe"
INFO: app cmdline: "seg_unittest.exe"
INFO: logdir is "C:\Users\SHill\AppData\Roaming\Dr. Memory"
INFO: symcache_dir is "C:\Users\SHill\AppData\Roaming\Dr. Memory\symcache"
INFO: Setting TZ to GMT Standard Time for i#2164 workaround
INFO: DynamoRIO configuration directory is C:\Users\SHill/dynamorio
INFO: configuring seg_unittest.exe pid=18496 dr_ops="-disable_traces -bb_single_restore_prefix -max_bb_instrs 256 -vm_size 256M -no_enable_reset -no_vm_base_near_app -no_early_inject -msgbox_mask 0 -logdir `C:\Users\SHill\AppData\Roaming\Dr. Memory\dynamorio` "
INFO: configuring client "C:\Program Files (x86)\Dr. Memory\bin64\release\drmemorylib.dll" ops="-logdir `C:\Users\SHill\AppData\Roaming\Dr. Memory` -symcache_dir `C:\Users\SHill\AppData\Roaming\Dr. Memory\symcache` -lib_blacklist `C:\Windows*.d??,C:\Program Files\Common Files\Microsoft Shared*.d??,C:\Program Files (x86)\Common Files\Microsoft Shared*.d??` -resfile 18496 "
INFO: waiting for app to exit...
~~Dr.M~~ Dr. Memory version 2.1.0
~~Dr.M~~ Running "seg_unittest.exe"
~~Dr.M~~ Using system call file C:\Program Files (x86)\Dr. Memory\bin64\syscalls_x64.txt
[...]

PR #2227, opened 2019-09-21
i#1438 MacOS: Fix crash with zero callstack frames

derekbruening:
Updates DR to 704b750f to fix memquery performance and
release build warnings (this was somehow omitted from a9800c4).

Fixes a crash that shows up with -brief on Mac64 but could happen anywhere:
callstack_next_retaddr() (called from check_type_match()) finds no frames
but tries to access them anyway, resulting in a null pointer deref.

Issue: #1438


derekbruening:
DrM failure is #2184

Issue #1925, opened 2016-11-13
drstrace_unit_tests is failing relatively consistently


derekbruening:
The drstrace_unit_tests test has been failing for a few weeks now in all
builds on a fairly regular basis:

```
"drstrace_unit_tests" start time: Oct 31 03:31 Eastern Daylight Time
Output:
----------------------------------------------------------
drfront_fetch_module_symbols failed 1 for |D:/derek/drmemory/nightly/run/build_drmemory-rel-32/bin/symfetch.dll| in dir ||
<end of output>
Test time =  37.42 sec
```

Xref #1612: drstrace module fetch fails when symfetch.dll is built with VS2013

When I try to repro, It's non-det: sometimes works, sometimes fails:

```
Testing symbol library fetching features
Using symbol path D:\derek\drmemory\git\build_x86_dbg\logs\drstrace_unit_tests_6096\symbols as the local store
Using symbol path srv*D:\derek\drmemory\git\build_x86_dbg\logs\drstrace_unit_tests_6096\symbols*http://msdl.microsoft.com/download/symbols to fetch symbols
drfront_fetch_module_symbols for |D:/derek/drmemory/git/build_x86_dbg/bin/symfetch.dll| in dir |D:\derek\drmemory\git\build_x86_dbg\logs\drstrace_unit_tests_6096|
  failed to fetch pdb for D:/derek/drmemory/git/build_x86_dbg/bin/symfetch.dll, got SymType 0
drfront_fetch_module_symbols failed 1 for |D:/derek/drmemory/git/build_x86_dbg/bin/symfetch.dll| in dir ||
```
 
   SymNone = 0,

Maybe a network issue?  Xref MS symbol server failures recently with win10 where a 2nd try works.  My theory is that this is a MS symbol server problem and nothing to do with DrM code or this particular hardware.  Maybe we should have the test just try again if the fetch fails?

derekbruening:
cf5c840553fcbe006dd21efa6b99d4b04c4924a3 put in 4 retries which is enough in manual testing and for most suite runs but there was still a failure several days later:

http://dynamorio.org/CDash/testDetails.php?test=153178&build=14669
drfront_fetch_module_symbols failed 1 for |D:/derek/drmemory/nightly/run/build_drmemory-rel-32/bin/symfetch.dll|

derekbruening:
http://dynamorio.org/CDash/testDetails.php?test=153786&build=14725
[...]

PR #2229, opened 2019-10-11
DRi#3886: Fix Catalina library segment alignment issue

derekbruening:
Pulls in DR 0720551f for a fix to DRi#3886 where library segments have
sub-page alignment and are packed onto the same page on Catalina.

Updates the min supported Mac kernel to 19 (Catalina).


derekbruening:
Appveyor is drstrace_unit_tests #1925 and the nondet crashes plaguing us there on x64
Travis failure: #1712

PR #2240, opened 2020-01-18
Upgrade Travis from Trusty to Xenial

derekbruening:
We're hitting issues with the old compiler versions on Trusty.
This upgrades to Xenial which matches DR.


derekbruening:
Travis failure is nondet malloc #1712

Issue #2180, opened 2019-03-17
Fix and un-ignore 64-bit Windows full mode failing tests


derekbruening:
Xref #1938 which covers the 32-bit tests.
Xref #111.  Full mode is now the default but we have not had time to fix all the tests that fail with that in place.  We added some to the ignore list to get Appveyor green and keep things moving along.  This issue covers going through the ignore list, fixing each test, and removing it from the list.

derekbruening:
These are the failures on my win10 machine, for reference:

```
85% tests passed, 17 tests failed out of 114
The following tests FAILED:
          6 - badjmp (Failed)
         19 - procterm (Failed)
         40 - suppress (Failed)
         41 - suppress-genoffs (Failed)
         42 - suppress-gensyms (Failed)
         44 - wincrt (Failed)
         53 - gdi (Failed)
         54 - syscalls_win (Failed)
         55 - handle (Failed)
         66 - nudge (Failed)
         75 - syscall_file_all (Failed)
         76 - syscall_file_gen (Failed)
         96 - drsyscall_test (Failed)
         97 - strace_test (Failed)
        109 - drstrace (Failed)
        112 - drltrace_symargs (Failed)
        114 - strace_sample (Failed)
Errors while running CTest
```

```
$ egrep 'Testing:|Timeout|\*Failed|, found|  found|ASSERT|failed to|internal crash|RIO usage error' Testing/Temporary/LastTest.log
19/114 Testing: procterm
  stderr failed to match " 0 unique, 0 total, 0 byte\(s\) of leak\(s\)"
6/114 Testing: badjmp
  failed to match "badjmp\.c:76"
40/114 Testing: suppress
  stderr failed to match "~~Dr\.M~~ 0 unique, 0 total unaddressable
  access\(es\)", found "~~Dr.M~~ 2 unique, 2 total unaddressable access(es)"
41/114 Testing: suppress-genoffs
  failed to match "Error #2: UNINITIALIZED READ: reading register"
42/114 Testing: suppress-gensyms
  failed to match "Error #2: UNINITIALIZED READ: reading register"
44/114 Testing: wincrt
  stderr failed to match "~~Dr\.M~~ 2 unique, 2 total invalid heap
  argument\(s\)", found "~~Dr.M~~ 0 unique, 0 total invalid heap argument(s)"
53/114 Testing: gdi
  stderr failed to match "~~Dr\.M~~ 12 unique, 12 total GDI usage
[...]

PR #2130, opened 2018-10-26
i#2283: Adds faulty redzones

johnfxgalea:
Introduces a new option that renders redzones as faulty for umbra_32. When an access, a read or write, is made on a redzone, a fault occurs. 

This is useful for detecting cross boundary accesses without the need to perform explicit checks. 

This functionality is similar to faults upon writes to shared blocks.

I'll write tests soon. I want to get comments on its API, and even more so, ensure that this functionality is wanted. I found it useful/required for my projects.

Issue #2283

PR #2282, opened 2020-03-16
i#2283: Adds a method for retrieving granularity info about an Umbra Map.

johnfxgalea:
Adds a new function umbra_get_granularity() which returns information about the granularity of an Umbra Map.

Issue: #2283

### DIFF
--- a/tests/framework/umbra_client_allscales.c
+++ b/tests/framework/umbra_client_allscales.c
@@ -46,7 +46,7 @@
 
 static void
 test_umbra_mapping(client_id_t id, umbra_map_scale_t scale, const char *label,
-                   const int scale_val, const bool is_scale_down)
+                   int scale_val, bool is_scale_down)
 {
     dr_printf("\n====================\ntesting scale %d == %s\n", scale, label);
     umbra_map_t *umbra_map;
--- a/tests/framework/umbra_client_allscales.c
+++ b/tests/framework/umbra_client_allscales.c
@@ -63,7 +63,7 @@ test_umbra_mapping(client_id_t id, umbra_map_scale_t scale, const char *label,
         CHECK(false, "failed to init umbra");
     if (umbra_create_mapping(&umbra_map_ops, &umbra_map) != DRMF_SUCCESS)
         CHECK(false, "failed to create shadow memory mapping");
-    if (umbra_get_granularity(umbra_map, &scale_val_out,&is_scale_down_out) != DRMF_SUCCESS)
+    if (umbra_get_granularity(umbra_map, &scale_val_out, &is_scale_down_out) != DRMF_SUCCESS)
             CHECK(false, "failed to get granularity info umbra");
     CHECK(scale_val == scale_val_out, "incorrect scale");
     CHECK(is_scale_down == is_scale_down_out, "incorrect scale granularity");
--- a/tests/framework/umbra_client_allscales.c
+++ b/tests/framework/umbra_client_allscales.c
@@ -45,10 +45,13 @@
                 __FILE__, __LINE__, #cond, msg), dr_abort(), 0)))
 
 static void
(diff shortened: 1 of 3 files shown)


umbra/umbra.c
```diff
@@ -758,3 +758,35 @@ umbra_get_shared_shadow_block(IN  umbra_map_t *map,
         return DRMF_ERROR_INVALID_PARAMETER;
     return umbra_get_shared_shadow_block_arch(map, value, value_size, block);
 }
+
+DR_EXPORT
+drmf_status_t umbra_get_granularity(const umbra_map_t *map, OUT int *scale,
```
derekbruening: style: newline between type and name

umbra/umbra.c
```diff
@@ -758,3 +758,35 @@ umbra_get_shared_shadow_block(IN  umbra_map_t *map,
...
+                                    OUT bool *is_scale_down) {
+
+  if (map == NULL || scale == NULL || is_scale_down == NULL)
+    return DRMF_ERROR_INVALID_PARAMETER;
```
derekbruening: style: every line in the function: indent 4

(I do want to get clang-format here; have not gotten to it yet)

umbra/umbra.h
```diff
@@ -725,6 +725,18 @@ umbra_shadow_memory_info_init(umbra_shadow_memory_info_t *info)
...
+DR_EXPORT
+/**
+ * Returns granularity information of the passed Umbra map.
+ *
+ * @param[in]  map              The mapping object to use.
+ * @param[out] scale            The pointer where to store the returned scale.
+ * @param[out] is_scale_down    The pointer where to store the returned flag
+ * indicating whether shadow memory is scaled down or up.
+ */
+drmf_status_t umbra_get_granularity(const umbra_map_t *map, OUT int *scale,
```
derekbruening: Is there a test in tests/framework/umbra_client_*.c where we could add a call to this function?

johnfxgalea: That is on my list b/c it is also blocking #2130

umbra/umbra.c
johnfxgalea: Done.

umbra/umbra.c
johnfxgalea: Done.

umbra/umbra.h
derekbruening: If you could clarify: Did you want to merge this as-is and put the test in separately, or you're going to wait for more tests to be added and then re-activate this PR?

johnfxgalea: I think it is best that we wait for tests prior to merge. I still need to see what tests exist for Umbra.

derekbruening: There are a few at tests/framework/umbra_client_*.c, one of which I added recently, but I am pretty sure they are not covering the whole API.

johnfxgalea: Should be ready now.

tests/framework/umbra_client_allscales.c
```diff
@@ -45,10 +45,13 @@
                 __FILE__, __LINE__, #cond, msg), dr_abort(), 0)))
 
 static void
-test_umbra_mapping(client_id_t id, umbra_map_scale_t scale, const char *label)
+test_umbra_mapping(client_id_t id, umbra_map_scale_t scale, const char *label,
+                   const int scale_val, const bool is_scale_down)
```
derekbruening: Not a fan of marking int and bool parameters as const since it has no impact on the caller or interface.  I guess our style guide doesn't say anything, but we're not doing this anywhere else, so for consistency I would suggest removing.  Maybe we should add it to the style guide.

derekbruening: I added it to the style page.

johnfxgalea: Done

johnfxgalea:
Failures due to #2180

Issue #1782, opened 2015-09-30
64-bit shadow mapping init fails on Fedora 22


derekbruening:
Many tests fail on x64 on Fedora 22:

```
drmemory-dbg-64: 21 tests passed, **** 8 tests failed: ****
```

Failures are due to shadow init failing:

```
        free.exitcode => ~~Dr.M~~ ASSERT FAILURE (thread 17833): /work/drmemory/git/src/drmemory/shadow.c:277: false (fail to create shadow memory mapping)

                CMake Error at runtest.cmake:311 (message):;  ~~Dr.M~~ ASSERT FAILURE (thread 17833):
                  /work/drmemory/git/src/drmemory/shadow.c:277: false (fail to create shadow
                  memory mapping)
        fuzz_buffer => ~~Dr.M~~ ASSERT FAILURE (thread 17837): /work/drmemory/git/src/drmemory/shadow.c:277: false (fail to create shadow memory mapping)

                CMake Error at runtest.cmake:311 (message):;  ~~Dr.M~~ ASSERT FAILURE (thread 17837):
                  /work/drmemory/git/src/drmemory/shadow.c:277: false (fail to create shadow
                  memory mapping)
```

Running

```
# bin64/drmemory -- tests/free.pattern
```

I can reproduce about 1/2 of the time.

Here is an example:

This fails:

```
                segment_overlap(base, end,
                                app_segments[i].reserve_base[map_idx],
                                app_segments[i].reserve_end[map_idx])) {
```

For:

0x00000340'00000000-0x00000380'00000000
0x00000370'00000000-0x0000037f'fff00000

  scale = UMBRA_MAP_SCALE_DOWN_4X, 

```
> cat /proc/29352/maps
00400000-00401000 r-xp 00000000 fd:09 15222034                           /work/drmemory/git/build_x64_dbg/tests/free.pattern
00401000-00600000 ---p 00000000 00:00 0 
00600000-00602000 rw-p 00000000 fd:09 15222034                           /work/drmemory/git/build_x64_dbg/tests/free.pattern
00602000-00603000 ---p 00000000 00:00 0 
00603000-00604000 rw-p 00000000 00:00 0 
51c94000-51c95000 ---p 00000000 00:00 0 
51c95000-51c9b000 rw-p 00000000 00:00 0 
51c9b000-51c9d000 ---p 00000000 00:00 0 
51c9d000-51c9e000 rw-p 00000000 00:00 0 
[...]

PR #2286, opened 2020-03-21
i#2285: Update DR and drcontainers for compatibility change

derekbruening:
Updates DR to 1d711b47 and addresses a source compatibility change in
hashtable_config_t by setting the structure to zeroes.

Issue: #2285


derekbruening:
Travis failure is #1782.

Issue #2250, opened 2020-02-09
Add VS2017 build support and update Appveyor to VS2017/Win10


derekbruening:
DR is now using VS2017 and it is complicating making a combined package (
https://github.com/DynamoRIO/dynamorio/issues/2861).

Plus, Appveyor with VS2013 is basically Win8.1.  Updating our Appveyor image to VS2017 will come with Win10 (1607) which will be a better test plaform.

derekbruening:
Pasting in my notes showing why PR #2265 wasn't merged yet:

## TODO list of test failures

Wow, a lot:
```
====> FAILURE in drmemory-dbg-32 <====
drmemory-dbg-32: 109 tests passed, **** 33 tests failed, but ignoring 5 for i1938: ****
	(ignore: i#1938) 	procterm 
	operators 
	state 
	procterm.nativeparent 
	(ignore: i#1938) 	winthreads 
	wrap_malloc 
	wrap_cs2bug 
	wrap_operators 
	(ignore: i#1938) 	wrap_wincrt 
	wincrt 
	wrap_wincrtdbg 
	mallocMD 
	cs2bugMTd 
	cs2bugMTdZI 
	cs2bugMD 
	wrap_cs2bugMTd 
	cs2bugMDd 
	operatorsMDd 
	gdi 
	handle 
	wrap_operatorsMDd 
	handle_only 
	blacklist 
	nudge 
	leak_string 
	nosyms 
	(ignore: i#1938) 	app_suite.pattern 
	syscall_file_all 
	(ignore: i#1938) 	app_suite 
	syscall_file_gen 
	drsyscall_test => ASSERT FAILURE: ..\..\tests\framework\drsyscall_client.c:206: strcmp(name, "NtQueryValueKey") == 0 (syscalls in this app shouldn't fail)
		</Value>
						%</Measurement>
	strace_test 
	drstrace_unit_tests 
WARNING: maximum warning/error limit hit for drmemory-dbg-64!
  Manually verify whether it succeeded.

====> FAILURE in drmemory-dbg-64 <====
drmemory-dbg-64: 29 tests passed, **** 86 tests failed, but ignoring 31 for i1938: ****
	hello 
	(ignore: i#1938) 	procterm 
	malloc 
	leak_indirect 
	free 
	(ignore: i#1938) 	badjmp 
	registers 
	asmtest 
	multierror 
	bitfield 
	bitfieldGL 
	(ignore: i#1938) 	cs2bug 
[...]

Issue #1798, opened 2015-10-22
Enable Umbra to proactively track memory allocation


zhaoqin:
Umbra is the module to maintain the shadow memory mapping for application memory. Currently, Umbra does not track the memory allocation but wait for the client calling shadow memory creation routines to create shadow memory. Umbra should be able to proactively track the allocation and manage the shadow memory accordingly. The challenges to the proactively tracking is that there are not only `mmap` and `NtAllocateVirtualMemory` that allocates memory from the OS, there are other system calls.
One option is to use `DrMalloc` (xref #824).

derekbruening:
We also want to prevent a tool's mappings from conflicting.  We want to put Umbra logic at 3 control points: DR placement (on Linux at least, by default today we can reload DR at inject time), priv lib placement, and app mmap placement.

derekbruening:
We were hoping that this could help ensure umbra's assumption of nothing being mapped in the 0x7fff'ff4-0x7fff'ff8 gap.

PR #2300, opened 2020-06-19
i#2283: Facilitate Umbra Layouts with Constraint Solving.

johnfxgalea:
Adds a helper python script to facilitate the setting up of Umbra Shadow Memory Layouts. Given the details of a memory layout, such as mask and disp values, the script will check whether any collisions are present with respect to the current pre-defined app regions that Umbra 64-bit considers today. 

**Example:**

```
python3 umbra_layout.py --os linux --mask 0xFFFFFFFFFFF --disp 0x120000000000
Information:
	OS: linux
SUCCESS

Memory Layout:

	Disp: 0x0000120000000000
	Mask: 0x00000FFFFFFFFFFF
	Unit Size: 0x0000100000000000
	Scale: 0
	Map Count: 1 

MAP 0
	app0:	 [0x0000000000000000, 0x0000010000000000] : exec,heap, data
	shd0:	 [0x0000120000000000, 0x0000130000000000]


	app1:	 [0x0000550000000000, 0x0000570000000000] : pie
	shd1:	 [0x0000170000000000, 0x0000190000000000]


	app2:	 [0x00007F0000000000, 0x0000800000000000] : lib, map, stack, vdso
	shd2:	 [0x0000210000000000, 0x0000220000000000]
```

The script also uses a constraint solver (namely Z3) to check a layout as well as synthesize satisfying displacement values that result in no collisions (if sat).

**Example:**

```
python3 umbra_layout.py --os linux --scale up_2x --mask 0xFFFFFFFFFFF --verify
Information:
	OS: linux
SUCCESS

Memory Layout:

	Disp: 0x0000140000000000
	Mask: 0x00000FFFFFFFFFFF
	Unit Size: 0x0000100000000000
	Scale: 1
	Map Count: 1 

MAP 0
	app0:	 [0x0000000000000000, 0x0000010000000000] : exec,heap, data
	shd0:	 [0x0000280000000000, 0x00002A0000000000]


	app1:	 [0x0000550000000000, 0x0000570000000000] : pie
	shd1:	 [0x0000320000000000, 0x0000360000000000]


	app2:	 [0x00007F0000000000, 0x0000800000000000] : lib, map, stack, vdso
	shd2:	 [0x0000460000000000, 0x0000480000000000]
```


derekbruening:
I was going to say, don't we already have some of this: but it looks like the app I have is not checked in anywhere.  It doesn't have a sat checker (used manual loops in the past).  It does, however, check all the scales (up 2x, down 2x, down 4x, down 8x) for any particular setting, if you could add that here.  I put it at #2301 (not to commit it just for comparison).

The checked-in test we have is umbra_client_allscales but it only checks the checked-in config.

johnfxgalea:
> It does, however, check all the scales (up 2x, down 2x, down 4x, down 8x) for any particular setting.

So the script already allows one to check a layout for a given scale. I can add support to the script to check all scales but I wonder what the benefit is because Umbra associates an individual layout setting for each scale.

derekbruening:
> > It does, however, check all the scales (up 2x, down 2x, down 4x, down 8x) for any particular setting.
> 
> So the script already allows one to check a layout for a given scale. I can add support to the script to check all scales but I wonder what the benefit is because Umbra associates an individual layout setting for each scale.

That's not how I remember Umbra working: it tries to use the same setting for all scales, and only grudgingly makes an exception with a different displacement, having to do it just for one (or 2?) of the scales.  Ideally there would be no exceptions, so I would want to see the results for all scales at once, preferring solutions with no exceptions.

johnfxgalea:
> > > It does, however, check all the scales (up 2x, down 2x, down 4x, down 8x) for any particular setting.
> > 
> > 
> > So the script already allows one to check a layout for a given scale. I can add support to the script to check all scales but I wonder what the benefit is because Umbra associates an individual layout setting for each scale.
> 
> That's not how I remember Umbra working: it tries to use the same setting for all scales, and only grudgingly makes an exception with a different displacement, having to do it just for one (or 2?) of the scales. Ideally there would be no exceptions, so I would want to see the results for all scales at once, preferring solutions with no exceptions.

Okay, no problem. I added the new option.

I think the implementation tries the same setting for all scales by design, but on Linux, a different displacement value is used for each scale:
https://github.com/DynamoRIO/drmemory/blob/2d1e0a51327ef96d075c2c41a3126f3ee262e765/umbra/umbra_64.c#L329

Interestingly, the script manages to synthesize a displacement value that should be suitable for all scales, assuming only one shadow map is registered.

```
python3 umbra_layout.py --os linux --mask 0xFFFFFFFFFFF --verify --scale all --count 1
*** Umbra Shadow Memory Layout ***

OS: linux 

SUCCESS
Memory Layout:
	Mask: 0x00000FFFFFFFFFFF
	Disp: 0x0000080000000000
	Unit Size: 0x0000100000000000
	Scale UP: False
	Scale: 3


MAP 0
	app0:	 [0x0000000000000000, 0x0000010000000000] : exec,heap, data
	shd0:	 [0x0000010000000000, 0x0000012000000000]


	app1:	 [0x0000550000000000, 0x0000570000000000] : pie
	shd1:	 [0x000001A000000000, 0x000001E000000000]


	app2:	 [0x00007F0000000000, 0x0000800000000000] : lib, map, stack, vdso
	shd2:	 [0x000002E000000000, 0x0000030000000000]


Memory Layout:
	Mask: 0x00000FFFFFFFFFFF
	Disp: 0x0000080000000000
	Unit Size: 0x0000100000000000
	Scale UP: False
	Scale: 2


MAP 0
	app0:	 [0x0000000000000000, 0x0000010000000000] : exec,heap, data
	shd0:	 [0x0000020000000000, 0x0000024000000000]


	app1:	 [0x0000550000000000, 0x0000570000000000] : pie
	shd1:	 [0x0000034000000000, 0x000003C000000000]


	app2:	 [0x00007F0000000000, 0x0000800000000000] : lib, map, stack, vdso
	shd2:	 [0x000005C000000000, 0x0000060000000000]


Memory Layout:
	Mask: 0x00000FFFFFFFFFFF
	Disp: 0x0000080000000000
	Unit Size: 0x0000100000000000
	Scale UP: False
	Scale: 1


MAP 0
	app0:	 [0x0000000000000000, 0x0000010000000000] : exec,heap, data
	shd0:	 [0x0000040000000000, 0x0000048000000000]


	app1:	 [0x0000550000000000, 0x0000570000000000] : pie
	shd1:	 [0x0000068000000000, 0x0000078000000000]


	app2:	 [0x00007F0000000000, 0x0000800000000000] : lib, map, stack, vdso
	shd2:	 [0x00000B8000000000, 0x00000C0000000000]


Memory Layout:
	Mask: 0x00000FFFFFFFFFFF
	Disp: 0x0000080000000000
	Unit Size: 0x0000100000000000
	Scale UP: True
	Scale: 0


MAP 0
	app0:	 [0x0000000000000000, 0x0000010000000000] : exec,heap, data
	shd0:	 [0x0000080000000000, 0x0000090000000000]


	app1:	 [0x0000550000000000, 0x0000570000000000] : pie
	shd1:	 [0x00000D0000000000, 0x00000F0000000000]


	app2:	 [0x00007F0000000000, 0x0000800000000000] : lib, map, stack, vdso
	shd2:	 [0x0000170000000000, 0x0000180000000000]


Memory Layout:
	Mask: 0x00000FFFFFFFFFFF
	Disp: 0x0000080000000000
	Unit Size: 0x0000100000000000
	Scale UP: True
	Scale: 1


MAP 0
	app0:	 [0x0000000000000000, 0x0000010000000000] : exec,heap, data
	shd0:	 [0x0000100000000000, 0x0000120000000000]


	app1:	 [0x0000550000000000, 0x0000570000000000] : pie
	shd1:	 [0x00001A0000000000, 0x00001E0000000000]


	app2:	 [0x00007F0000000000, 0x0000800000000000] : lib, map, stack, vdso
	shd2:	 [0x00002E0000000000, 0x0000300000000000]


Memory Layout:
	Mask: 0x00000FFFFFFFFFFF
	Disp: 0x0000080000000000
	Unit Size: 0x0000100000000000
	Scale UP: True
	Scale: 2


MAP 0
	app0:	 [0x0000000000000000, 0x0000010000000000] : exec,heap, data
	shd0:	 [0x0000200000000000, 0x0000240000000000]


	app1:	 [0x0000550000000000, 0x0000570000000000] : pie
	shd1:	 [0x0000340000000000, 0x00003C0000000000]


	app2:	 [0x00007F0000000000, 0x0000800000000000] : lib, map, stack, vdso
	shd2:	 [0x00005C0000000000, 0x0000600000000000]


Memory Layout:
	Mask: 0x00000FFFFFFFFFFF
	Disp: 0x0000080000000000
	Unit Size: 0x0000100000000000
	Scale UP: True
	Scale: 3


MAP 0
	app0:	 [0x0000000000000000, 0x0000010000000000] : exec,heap, data
	shd0:	 [0x0000400000000000, 0x0000480000000000]


	app1:	 [0x0000550000000000, 0x0000570000000000] : pie
	shd1:	 [0x0000680000000000, 0x0000780000000000]


	app2:	 [0x00007F0000000000, 0x0000800000000000] : lib, map, stack, vdso
	shd2:	 [0x0000B80000000000, 0x0000C00000000000]

```
If I try to consider additional maps,  separated by a 2 units, a single setting for all scales seems to be unsat on Linux.

```
python3 umbra_layout.py --os linux --mask 0xFFFFFFFFFFF --verify --scale all --count 2
*** Umbra Shadow Memory Layout ***

OS: linux 

FAILED
```

derekbruening:
@johnfxgalea -- is this mergeable?  It may be very useful for #2328 where shadow layout problems are blocking us.

johnfxgalea:
Yep, it is ready to be merged but still awaiting a review I think.

derekbruening:
> Yep, it is ready to be merged but still awaiting a review I think.

Oh, whoops.  Though no longer sure #2328 is really a layout issue: may instead be an initialization/address walk issue at init time.

tools/umbra_layout.py
```diff
@@ -0,0 +1,566 @@
...
+        # Returns pre-defined app regions of the passed Operating System.
+        regions = []
+        if os == OS.LINUX:
+            regions.append(Region('exec,heap, data', 0x0, 0x10000000000))
+            regions.append(Region('pie', 0x550000000000, 0x570000000000))
+            regions.append(Region('lib, map, stack, vdso', 0x7F0000000000,
+                                  0x800000000000))
+
+            # FIXME: Should we map shadow memory for vsyscall?
+            # Doing so can prevent the possibility of a SAT layout.
```
derekbruening, line 69: But we have to, and the current C code handles vsyscall.  Are you saying there's only code and no data and nobody shadows code?

tools/umbra_layout.py
```diff
@@ -0,0 +1,566 @@
...
+            return True
+        elif (scale == Scale.UP_4X):
+            return True
+        elif (scale == Scale.UP_8X):
+            return True
+        else:
+            sys.exit('Fatal Error: Unknown scale.')
+
+    def get_scale(scale):
+        # Returns whether the scale's value.
```
derekbruening, line 122: s/whether//

tools/umbra_layout.py
```diff
@@ -0,0 +1,566 @@
...
+        if self.is_scale_up:
+            unit_disp = unit_disp >> self.scale
+        else:
+            unit_disp = unit_disp << self.scale
+
+        # Step 3: Calculate the shadow address
+        shdw_addr = masked_addr + self.disp
+
+        # Step 4: Handle special case where top of addr is masked out, e.g., 0x800000000.
+        if addr != 0 and shdw_addr == self.disp:
```
derekbruening, line 203: Not understanding this: this step would not be there for current umbra code??  Could you elaborate on this?

johnfxgalea: That models this piece of code:

https://github.com/DynamoRIO/drmemory/blob/4fdcb814500a62317bf5659bba6a05a900c01dc9/umbra/umbra_64.c#L424

tools/umbra_layout.py
johnfxgalea: Yeah, back then I was thinking of whether we need to have shadow for code. However, "we have" to because we want to catch erroneous use of memory right?

derekbruening: Right, I think we have to handle all mapped app memory regions.

tools/umbra_layout.py
derekbruening: Woah, still not understanding: inlined shadow translation code certainly does not have a branch and so this differs from what's inlined, so I'm not sure how it can work?  @zhaoqin do you recognize this?

johnfxgalea: Yea.... I agree with your reading regarding inlining.

Basically, what the code is doing here is handling the case where the mask clears the top part of an address, which is essentially the data entropy required to set up a layout. 

if this condition, namely addr == map->disp, is confusing you, take a look at the previous line, where the addr is computed. If the mask zeros out the pc, then addr would be equal to map->disp.

https://github.com/DynamoRIO/drmemory/blob/4fdcb814500a62317bf5659bba6a05a900c01dc9/umbra/umbra_64.c#L423

derekbruening: > If the mask zeros out the pc, then addr would be equal to map->disp.

Right, but that is expected: that's just part of the "mask and add" approach that you have to take into account when setting up the masks and disps.  If the pc is `0x800'00000001` we get disp+1 for addr; if the pc is one less we get disp.  Not sure *why* it tries to handle just the equals-disp part here and why it's so different from other cases.

The only thing I can think of is that it's trying to handle open endpoints.  So if this is our app region and it does not include 0x800':
```
 *   app2: [0x000007F0'00000000, 0x00000800'00000000): lib
```

Is it trying to handle someone iterating over that range and (incorrectly) *including* the endpoint or something, like someone printing out the endpoint of the corresponding shadow region?  This is something I've run into actually: line 51 at https://github.com/DynamoRIO/drmemory/pull/2301/files#diff-e88a4a2cc5d58ef58b542c5ec5473ac798f340de617d3b32bfc157fec6494566 where I subtract one from the end before translating to shadow, and then add 1 back on.

johnfxgalea: Hmm, I think you are on to something there. That could explain why this special case is not handled in inlined code.

Should I, therefore, remove this special case from umbra_layout.py to reflect inlined code?

derekbruening: Yes, I think it should be removed from the .py, and also removed from here with instead a tweak to whatever callers are trying to find shadow endpoints to have them do the -1,+1 solution -- but since that takes some work, esp to verify with the lack of good umbra tests, maybe in this PR just add a comment to the umbra_64.c code?

derekbruening: I talked to @zhaoqin and he thinks it is indeed for the endpoint of a shadow block, and the -1,+1 solution in the callers is the alternative.

johnfxgalea: Awesome, thanks for checking!

derekbruening: I filed #2330 on the cleanup in the umbra code.

derekbruening: Maybe as part of this PR you could put in a TODO i#2330 comment in the umbra code.

johnfxgalea: Done.

tools/umbra_layout.py
johnfxgalea: Done.

tools/umbra_layout.py
johnfxgalea: Done.

tools/umbra_layout.py
johnfxgalea: Let me know if this is okay. Hit resolve if it is fine.

johnfxgalea:
Done. PTAL. I added some further comments.

tools/umbra_layout.py
derekbruening: LGTM.

derekbruening (approved):
(Unfotunately the CI has been in bad shape for a while...I moved unix to Github Actions, and am trying to move Windows there and make it green too, but there are many problems to resolve.)

tools/umbra_layout.py
johnfxgalea: Hmm, so I was doing some final checks before I go ahead with the merge, and I note that if we include vsyscall, then certain layouts are not possible.

Currently, on Linux, if we select scale x2 up, we get a collision. 

```
*** Umbra Shadow Memory Layout ***

OS: linux 

Memory Layout:
	Mask: 0x00000FFFFFFFFFFF
	Disp: 0x0000048000000000
	Unit Size: 0x0000100000000000
	Scale UP: True
	Scale: 1


Collision: 0x000028FFFEC00000 0x000028FFFFFFFFFE
FAILED

MAP 0
	app0:	 [0x0000000000000000, 0x000000FFFFFFFFFF] : exec,heap, data
	shd0:	 [0x0000090000000000, 0x00000AFFFFFFFFFE]


	app1:	 [0x0000550000000000, 0x000055FFFFFFFFFF] : pie
	shd1:	 [0x0000130000000000, 0x000014FFFFFFFFFE]


	app2:	 [0x00007F0000000000, 0x00007FFFFFFFFFFF] : lib, map, stack, vdso
	shd2:	 [0x0000270000000000, 0x000028FFFFFFFFFE]


	app3:	 [0xFFFFFFFFFF600000, 0xFFFFFFFFFF600FFF] : vsyscall
	shd3:	 [0x000028FFFEC00000, 0x000028FFFEC01FFE]

```

It collides with lib region. This is the layout setting we use today. It also appears that the problem is only when we scale up shadow memory (and not scale down).

johnfxgalea: https://github.com/DynamoRIO/drmemory/blob/ef8db941a3db47e4a6778448be68f53a8db048a0/umbra/umbra_64.c#L314

johnfxgalea: If I try to get a solution from the model, I get unsat:

```
$ python3 umbra_layout.py --os linux --mask 0xFFFFFFFFFFF --scale up_2x --verify --count 1
*** Umbra Shadow Memory Layout ***

OS: linux 

FAILED

```

derekbruening: As your link above points at, the code today has a custom disp for up-2x.  Are you able to add varying disps per scale to your model?

johnfxgalea: Yeah, it does. The script can check fixed disps passed as param, or ask the solver to provide custom disps (if sat).

I checked that the disp currently used by Umbra here: https://github.com/DynamoRIO/drmemory/pull/2300#discussion_r537068124

I asked the solver to provide me with a custom disp (if sat),  here: https://github.com/DynamoRIO/drmemory/pull/2300#discussion_r537069155

But it's not sat.

derekbruening: So you're saying that the current Umbra code is incorrect.  But I thought there were sanity checks inside the umbra code itself, plus tests like umbra_all_scales: are there not enough checks there, or is there a bug in one of them?

johnfxgalea: Not a bug per se because Umbra code ignores vsyscall.

https://github.com/DynamoRIO/drmemory/blob/55a9c58b75ab5dbcabdb6368c78132d2eab6fbc1/umbra/umbra_64.c#L357

If however, we consider it in our model, then it is unsat. I guess the main point is that Umbra code is currently ignoring mapping of vsyscall and thus breaking its contract.

https://github.com/DynamoRIO/drmemory/pull/2300#discussion_r532875525

johnfxgalea: Perhap, we shoudl revert, adding again the FIXME comment?

derekbruening: I'm still confused: I don't see the Umbra code deliberately ignoring vsyscall?
For what you're quoting:
```
    /* We split app3 [0x7F0000000000, 0x800000000000) into two parts:
```
It goes on to say:
```
     * [0x7F0000000000, 0x7FFFFF400000) and [0x7FFFFF800000, 0x800000000000).
     * And we skip [0x7FFFFF400000-0x7FFFFF800000)
...
     * We assume [0x7FFFFF400000-0x7FFFFF800000) will not be used by app.
     * If app allocates memory from that region, umbra_add_app_segment
     * will fail because umbra_add_shadow_segment fails to add corresponding
     * shadow memory segment.
```

So if a vsyscall were to be in that skipped region, it would just fail up front.
Is that what you mean?  That's not great, but it's not what I thought you were
saying, which is that it pretending to have a correct mapping.

johnfxgalea: Hmm, basically, this thread concerns whether the layout should consider the shadow of vsyscall. Initially, the .py script did not consider this region and included a FIXME comment, highlighting that some layouts won't be possible otherwise. After some discussion, we removed this comment, and made the .py script to also take into account the vsyscall region.

However, the model does not skip any part of the app3. Consequently, a collision is detected.

I guess my mistake in explaining this was confusing the skipped region with vsyscall but I think my overall issue is still there nonetheless. By not considering mapping shadow to the skipped region, aren't we breaking Umbra's contract?

derekbruening: There are possible remediations: If the kernel isn't the one placing something in the gap at process init (like vsyscall) and it's the app who does an explicit mmap there or ld.so tries to load a library there, umbra can reject the mmap: so it's ok to have a gap if you can enforce it (though it's not today: #1798).  Or umbra could handle an unmovable alloc there by using expensive mappings: I think that's what #1782 proposes and the comment there implies.

derekbruening: I'm also confused by the vsyscall address in this PR: isn't vsyscall subject to ASLR and it moves around?

derekbruening: For the problematic app region where umbra_64.c adds a gap: wouldn't other tools who do simple displacement shadow mapping also hit a problem?  E.g., AddressSanitizer.

derekbruening: > I'm also confused by the vsyscall address in this PR: isn't vsyscall subject to ASLR and it moves around?

So if we're fairly certain vsyscall has nothing but code wrappers and no data, is not subject to ASLR so we know where it is, and is always separate from the larger and randomized vdso and vvar etc. which we are shadowing: then maybe skipping vsyscall is ok?

derekbruening:
Xref #1712: we're hitting umbra shadow mapping failures up front on Linux on Github Actions.

tools/umbra_layout.py
johnfxgalea: > I'm also confused by the vsyscall address in this PR: isn't vsyscall subject to ASLR and it moves around?

Isn't the location of vsyscall fixed? hence the introduction of vdso?

> For the problematic app region where umbra_64.c adds a gap: wouldn't other tools who do simple displacement shadow mapping also hit a problem? E.g., AddressSanitizer.

This is a good question. I do not know how Asan handles such issues, but it would be interesting to know. I'll try and search.

>  maybe skipping vsyscall is ok?

That is what I was thinking initially. I guess we need to decide whether we are going to change .py or Umbra code (i.e., skip a region in app3 or skip vsyscall), because we want the implementation to match the model.

@zhaoqin Any opinion on the above please?

johnfxgalea: I had some time to continue on this issue. I had a quick look at how ASAN sets up and maps shadow memory.

Doc:  https://github.com/google/sanitizers/wiki/AddressSanitizerAlgorithm
Src: https://code.woboq.org/llvm/llvm/lib/Transforms/Instrumentation/AddressSanitizer.cpp.html

According to ASAN's [mapping](https://code.woboq.org/llvm/compiler-rt/lib/asan/asan_mapping.h.html#21),  0x00007fff8000 - 0x10007fff7fff is reserved for shadow (and shadow's shadow).

```
// Typical shadow mapping on Linux/x86_64 with SHADOW_OFFSET == 0x00007fff8000:
--
// \|\| `[0x10007fff8000, 0x7fffffffffff]` \|\| HighMem    \|\|
// \|\| `[0x02008fff7000, 0x10007fff7fff]` \|\| HighShadow \|\|
// \|\| `[0x00008fff7000, 0x02008fff6fff]` \|\| ShadowGap  \|\|
// \|\| `[0x00007fff8000, 0x00008fff6fff]` \|\| LowShadow  \|\|
// \|\| `[0x000000000000, 0x00007fff7fff]` \|\| LowMem     \|\|
//
```

The shadow memory block is reserved upon initialization:  https://code.woboq.org/llvm/compiler-rt/lib/asan/asan_shadow_setup.cpp.html#126

Just for context, one thing to note is that ASAN maps an app byte to 1 bit - working at the granularity of 8 bytes. In contrast, Dr. Memory shadows two bits per byte, as denoted [here](https://github.com/DynamoRIO/drmemory/blob/04b22efd7e22f319c1c6e90f40186ec18500ad64/drmemory/shadow.c#L170).

johnfxgalea: Perhaps, we should do the model like how Umbra does it today, i.e., skip a region in app3. The issue would then be resolved with https://github.com/DynamoRIO/drmemory/issues/1798 where Umbra would enforce that this region is not used by the app.

johnfxgalea: @derekbruening I pushed the region split. Do you think this is good please?

derekbruening: Trying to context switch back in here: so this .py is now matching Umbra, and we just need to get Umbra to actually enforce its skip there.  Yes, SGTM.

derekbruening:
I thought I got Windows green -- are those flakes?  Add to the ignore list?

tools/umbra_layout.py
johnfxgalea: Great, thanks. We are on the same page.

johnfxgalea:
> I thought I got Windows green -- are those flakes? Add to the ignore list?

Does not seem to be a flake, with the following stable error:

`Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."
`

johnfxgalea:
> > I thought I got Windows green -- are those flakes? Add to the ignore list?
> 
> Does not seem to be a flake, with the following stable error:
> 
> `Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found." `

@derekbruening Any idea what the above error is about please? Happy to investigate further, but just wondering if you have any idea?

derekbruening:
> > > I thought I got Windows green -- are those flakes? Add to the ignore list?
> > 
> > 
> > Does not seem to be a flake, with the following stable error:
> > `Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found." `
> 
> @derekbruening Any idea what the above error is about please? Happy to investigate further, but just wondering if you have any idea?

If that's downloading doxygen is it related to https://github.com/DynamoRIO/dynamorio/issues/4654?

johnfxgalea:
> > > > I thought I got Windows green -- are those flakes? Add to the ignore list?
> > > 
> > > 
> > > Does not seem to be a flake, with the following stable error:
> > > `Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found." `
> > 
> > 
> > @derekbruening Any idea what the above error is about please? Happy to investigate further, but just wondering if you have any idea?
> 
> If that's downloading doxygen is it related to [DynamoRIO/dynamorio#4654](https://github.com/DynamoRIO/dynamorio/issues/4654)?

No. Dr Memory's C is making use of a stale URL (http://doxygen.nl/files/doxygen-1.8.19.windows.x64.bin.zip) here:

https://github.com/DynamoRIO/drmemory/blob/04b22efd7e22f319c1c6e90f40186ec18500ad64/.github/workflows/ci-windows.yml#L64

derekbruening:
> No. Dr Memory's CI is making use of a stale URL (http://doxygen.nl/files/doxygen-1.8.19.windows.x64.bin.zip) here:
> 
> https://github.com/DynamoRIO/drmemory/blob/04b22efd7e22f319c1c6e90f40186ec18500ad64/.github/workflows/ci-windows.yml#L64

Ah yeah it needs the equiv of PR #4643.

johnfxgalea:
CI failures concern  #2250 and #2341

This PR does not concern these issues.

Issue #2283, opened 2020-03-16
Bolster Umbra functionality


johnfxgalea:
This issue concerns a series of improvements to Umbra. In particular:
1. Add Umbra tests
2. Support additional granularities
3. Enable dynamic deletion of redundant blocks.

johnfxgalea:
We also have #2130 pending, which we will address once Umbra tests are in order.

johnfxgalea:
The shadow address of a shadow address is considered as an invalid address in Umbra.

However, Umbra has checks that ensure the shadow of an app region does not collide with the shadow's shadow when setting up memory as a form of double protection. The problem is that such checks block a shadow's shadow region from being legitimately used as a normal shadow region when we consider large scales, such as UP_4x or UP_8x. 

In other words, while current (low) scales can comply with such constraints in the layout, if we consider larger scale-ups, these constraints will prevent a successful layout.

Since shadow's shadow addresses are invalid, and the checks performed are more for debugging (rather than usage checks), we should remove the checks with the benefit of supporting for scales.

derekbruening:
> Since shadow's shadow addresses are invalid, and the checks performed are more for debugging (rather than usage checks), we should remove the checks with the benefit of supporting more scales.

This is not just for debugging: a shadow mapping scheme should map any address not allocated by the application to invalid memory.  Applications reference wild addresses on occasion and those should not map to application or shadow memory, causing tools to corrupt their own state or the application state.  This is especially bad for tools whose goal is to detect memory bugs in applications, such as Dr. Memory.

derekbruening:
If that invariant is dropped to make layout more flexible it needs to be a very explicit and documented decision and only tools that are ok with the reduced functionality should use it.

PR #2309, opened 2020-08-31
i2283: Adds more scales to Umbra

johnfxgalea:
Adds more scales to Umbra. In particular, the following scales are added:

- UMBRA_MAP_SCALE_DOWN_64X
- UMBRA_MAP_SCALE_DOWN_32X
- UMBRA_MAP_SCALE_UP_4X
- UMBRA_MAP_SCALE_UP_8X

I add support only for 32-bit. Support for 64-bit will be added via a separate PR.

PR #2310, opened 2020-09-01
i#2283: Remove shadow of shadows checks

johnfxgalea:
Removes rudimentary checks that block layouts for larger scales.

Issue: #2283

### DIFF
--- a/umbra/umbra_64.c
+++ b/umbra/umbra_64.c
@@ -49,9 +49,7 @@
  * and the result:
  *   shdw1 = SHDW(app1): [0x00000020'00000000, 0x00000030'00000000)
  *   shdw2 = SHDW(app2): [0x00000110'00000000, 0x00000120'00000000)
- * and
- *   shdw1'= SHDW(shdw1): [0x00000040'00000000, 0x00000050'00000000)
- *   shdw2'= SHDW(shdw2): [0x00000030'00000000, 0x00000040'00000000)
+ *
  * Here we call [0x00000000'00000000, 0x00000100'00000000) a unit, and each unit
  * has 16 (NUM_SEGMENTS = 0x100'00000000/0x10'00000000) segments
  * with size of 0x10'00000000.
@@ -66,9 +64,7 @@
  * and the result:
  *   shdw1 = SHDW(app1): [0x00000700'00000000, 0x00000a00'00000000)
  *   shdw2 = SHDW(app2): [0x00001300'00000000, 0x00001700'00000000)
- * and
- *   shdw1'= SHDW(shdw1): [0x00000e00'00000000, 0x00001100'00000000)
- *   shdw2'= SHDW(shdw2): [0x00000a00'00000000, 0x00000e00'00000000)
+ *
  * Here we call [0x00000000'00000000, 0x00001000'00000000) a unit, and each
  * unit has 16 segments with size of 0x100'00000000.  Note that we do allow
  * multi-segment regions.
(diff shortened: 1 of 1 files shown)


johnfxgalea:
See: https://github.com/DynamoRIO/drmemory/issues/2283#issuecomment-684765242

johnfxgalea:
Not sure if @zhaoqin would also like to also comment on the PR please.

umbra/umbra_64.c
```diff
@@ -523,37 +501,8 @@ umbra_add_shadow_segment(umbra_map_t *map, app_segment_t *seg)
...
                 ELOG(1, "ERROR: new app segment ["PFX", "PFX")'s shadow segment "
-                     "["PFX", "PFX") conflicts with app seg ["PFX", "PFX") or its "
-                     "shadow ["PFX", "PFX") or reserve ["PFX", "PFX")\n",
-                     seg->app_base, seg->app_end, base, end,
-                     app_segments[i].app_base, app_segments[i].app_end,
-                     app_segments[i].shadow_base[map_idx],
-                     app_segments[i].shadow_end[map_idx],
-                     app_segments[i].reserve_base[map_idx],
-                     app_segments[i].reserve_end[map_idx]);
-                return false;
```
derekbruening, line 539: Echoing https://github.com/DynamoRIO/drmemory/issues/2283#issuecomment-684894144: This is part of the core invariant of a shadow memory scheme.  A proper shadow scheme should map any address not allocated by the application to invalid memory. Applications reference wild addresses on occasion and those should not map to application or shadow memory, causing tools to corrupt their own state or the application state. This is especially bad for tools whose goal is to detect memory bugs in applications, such as Dr. Memory.

If we were to consider relaxing this invariant, it should only be done on the mapping schemes where there is no solution otherwise, and the lack of the invariant should be made very explicit and documented so that tools that care about it (such as Dr. Memory) can avoid it.

johnfxgalea: Thank you for raising this. So, I assumed that since UMBRA does not support meta-data for shadow regions, then checking whether a shadow's shadow collides can be relaxed because the user should never pass a shadow address for translation again. I guess a faulty app may inadvertently craft and use a shadow address due to undefined behaviour, which results in Dr Memory to translate this address again... But if the address can by anything, then why do we stop at checking whether the shadow's shadow collides with app/shadow regions, and not keep on going? i.e., check the shadow's shadow's shadow?

derekbruening: > But if the address can by anything, then why do we stop at checking whether the shadow's shadow collides with app/shadow regions, and not keep on going? i.e., check the shadow's shadow's shadow and so on?

Ideally nothing would map to the shadow memory except app memory, but I think the best we can do is to ensure that either the wild address or the shadow-mapped address is invalid and will fault.  At least that's what we put in our ["Efficient Memory Shadowing for 64-bit Architectures"](http://burningcutlery.com/derek/docs/ems64-ISMM10.pdf) paper.  That's why we focused on shadow-of-shadow, b/c the app ref won't fault.  So I guess you'd live with corruption of your shadow if you have your instrumentation prior to the app load of a wild address that does map to your shadow, and then when the app load faults you have to check in a fault handler and see that it messed you up and halt the program in case the app has its own handler that continues on wild faults?  @zhaoqin do you remember whether we had some model of post-instrumentation instead of pre-instrumentation or is it how I'm describing it here?

johnfxgalea: Okay, I understood. Thanks.

>  So I guess you'd live with corruption of your shadow if you have your instrumentation prior to the app load of a wild address that does map to your shadow, and then when the app load faults you have to check in a fault handler and see that it messed you up and halt the program in case the app has its own handler that continues on wild faults?

Yes. Basically what I am proposing is to remove reserved regions in order to make room for large scales. I know it is not ideal but haven't I thought of a better idea till now. 

>  then when the app load faults you have to check in a fault handler 

Actually, a fault might not occur at all; the app ref won't fault b/c it wildly addresses shadow memory, and the shadow's shadow address won't fault b/c it maps back to an app or shadow region as well. Hence corrupting the state

I guess I shouldn't have just removed the checks, but elided them only for larger scales as mentioned here:  https://github.com/DynamoRIO/drmemory/issues/2283#issuecomment-684895503

derekbruening: > Actually, a fault might not occur at all

To clarify: I was describing the original scheme where for each app load/store address A either 1) A is in a listed app region or 2) A is invalid (will fault) or 3) shadow(A) is invalid, and how it would interact with pre-load instru.

We're on the same page: removing the shadow-shadow check breaks the invariant and now a tool can corrupt its shadow or the app without (easily) detecting it, which should be acknowledged in the interface.

johnfxgalea:
Closed because this issue is to be resolved in https://github.com/DynamoRIO/drmemory/pull/2309

umbra/umbra_64.c
johnfxgalea: I'm closing this PR. Please see https://github.com/DynamoRIO/drmemory/pull/2309

Issue #2334, opened 2020-12-07
Move Windows CI to Github Actions


derekbruening:
Xref similar issue for DR: https://github.com/DynamoRIO/dynamorio/issues/4131

With the loss of Travis we have moved our unix CI to Github Actions: #2323.
This covers also moving our Windows CI from Appveyor, to gain parallel builds and auto-prior-build cancellation (https://github.com/DynamoRIO/dynamorio/issues/4585).

My first attempts are in PR #2326.  The GA image we're targeting is Server 2016, the same as our attempt to update our Appveyor image in #2250 and PR #2265.  Just like there, we hit major failures for 64-bit, due to #2328, which is now fixed; and major 32-bit failures: #2329.  Both were DR problems and both are now fixed.

Now we're doing to a slew of test failures:

32-bit:
https://pipelines.actions.githubusercontent.com/tMvJIh3ZIUXIfOP7dfgG7UBvQKftGfP2Bp1gAjRieRAnDKBVqy/_apis/pipelines/1/runs/102/signedlogcontent/4?urlExpires=2020-12-06T21%3A21%3A31.3518373Z&urlSigningMethod=HMACV1&urlSignature=UM25pzPPpL5cFdRg9YFc7W%2B%2B4HA%2BYGaWei0XZiag1Ds%3D
```
2020-12-06T16:51:28.5125095Z drmemory-dbg-32: 112 tests passed, **** 33 tests failed, but ignoring 4 for i1938: ****
2020-12-06T16:51:28.5125807Z 	procterm 
2020-12-06T16:51:28.5126266Z 	operators 
2020-12-06T16:51:28.5126722Z 	cs2bug 
2020-12-06T16:51:28.5127129Z 	state 
2020-12-06T16:51:28.5127781Z 	procterm.nativeparent 
2020-12-06T16:51:28.5128548Z 	malloc_callstacks 
2020-12-06T16:51:28.5129032Z 	reachable 
2020-12-06T16:51:28.5129567Z 	wrap_malloc 
2020-12-06T16:51:28.5130060Z 	wrap_operators 
2020-12-06T16:51:28.5130598Z 	(ignore: i#1938) 	wrap_cs2bug 
2020-12-06T16:51:28.5131082Z 	wrap_wincrt 
2020-12-06T16:51:28.5131529Z 	wincrt 
2020-12-06T16:51:28.5131977Z 	wrap_wincrtdbg 
[...]

PR #2335, opened 2020-12-07
Print contents of results file on CI test failures

derekbruening:
On a failure to match either the stderr summary or individual results,
if the env var CI is set (true for Github Actions), runtest.cmake now
prints out the contents of the results file(s).  This greatly aids
remote debugging of CI failures.

Issue: #2334


derekbruening:
x86 failures are #1712 shadow mapping failures: grrr.  But unrelated to this.

Issue #2333, opened 2020-12-07
Uninits in RtlCaptureContext in many tests on Server2016


derekbruening:
Extra uninits are present in many tests on Github Actions Server2016:
```
2020-12-07T02:55:07.8828198Z 4: Error #10: UNINITIALIZED READ: reading 0x00000036627fe2c0-0x00000036627fe320 96 byte(s) within 0x00000036627fe220-0x00000036627fe420
2020-12-07T02:55:07.8829450Z 4: # 0 ntdll.dll!RtlCaptureContext (0x00007ffe87809554 <ntdll.dll+0xa9554>)
2020-12-07T02:55:07.8830075Z 4:     ??:0
2020-12-07T02:55:07.8830696Z 4: # 1 ntdll.dll!RtlUnwindEx (0x00007ffe87795428 <ntdll.dll+0x35428>)
2020-12-07T02:55:07.8831255Z 4:     ??:0
2020-12-07T02:55:07.8831916Z 4: # 2 ntdll.dll!RtlImageNtHeaderEx (0x00007ffe877949d3 <ntdll.dll+0x349d3>)
2020-12-07T02:55:07.8832556Z 4:     ??:0
2020-12-07T02:55:07.8833151Z 4: # 3 __longjmp_internal (0x00007ff6e1cd263a <malloc.exe+0x1263a>)
2020-12-07T02:55:07.8833940Z 4:     d:\agent\_work\2\s\src\vctools\crt\vcruntime\src\eh\amd64\longjmp.asm:144
2020-12-07T02:55:07.8834749Z 4: # 4 __scrt_common_main_seh (0x00007ff6e1cc1918 <malloc.exe+0x1918>)
2020-12-07T02:55:07.8835584Z 4:     d:\agent\_work\2\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288
2020-12-07T02:55:07.8836690Z 4: # 5 our_top_handler (0x00007ff6e1cc102a <malloc.exe+0x102a>)
2020-12-07T02:55:07.8837624Z 4:     d:\a\drmemory\drmemory\tests\malloc.c:75
2020-12-07T02:55:07.8838763Z 4: # 6 KERNELBASE.dll!UnhandledExceptionFilter (0x00007ffe83ca6227 <KERNELBASE.dll+0x56227>)
2020-12-07T02:55:07.8839801Z 4:     ??:0
[...]

PR #2336, opened 2020-12-07
i#2333: False positives in RtlCaptureContext

derekbruening:
Adds RtlCaptureContext to the default Windows suppressions.  Errors
assumed to be false positives are showing up on Server 2016 and this
will help get our tests green.  We should investigate and improve our
CONTEXT handling to avoid needing this for a beter long-term solution.

Issue: #2333


derekbruening:
x86 failure is #1712 again.

Issue #2323, opened 2020-11-27
Migrate ASAP from Travis CI as it no longer offers free OSS CI and has stopped running


derekbruening:
See the full explanation in the corresponding DR issue:
https://github.com/DynamoRIO/dynamorio/issues/4549

The Travis account seems to be shared as the DrM Travis is now blocked even though it certainly hasn't used 10K credits on its own:
https://travis-ci.com/github/DynamoRIO/drmemory/pull_requests
> Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.

The plan is to move to Github Actions, just like for DR.

derekbruening:
I have the core UNIX CI set up now.  Packages, docs, cron jobs are not yet set up.

I tried Windows on the Server 2016 images and hit #2328 for 64-bit (same as on Appveyor: #2250; PR #2265) and for 32-bit every single test hung and timed out!  Grrr: not going to be easy to debug that.

derekbruening:
Filed #2329 for 32-bit.

PR #2343, opened 2020-12-12
i#2323 GA CI: Change trigger to support forked repos

derekbruening:
Updates the Github Actions CI trigger to limit pushes to master and
add pull request open, reopen, and synchronize events.  This supports
PR's from forked repositories without adding duplicate triggers.

Issue: #2323


derekbruening:
win32 has another failure to add to the ignore list: operatorsMDd

PR #2345, opened 2020-12-12
i#2323 GA CI: Add auto-cancellation of prior PR jobs

derekbruening:
Follows the lead of DynamoRIO/dynamorio#4585 and adds
auto-cancellation of a prior running job on a PR CI trigger.  Pushes
to master are never canceled.

Issue: #2323, DynamoRIO/dynamorio#4585

### DIFF
--- a/.github/workflows/ci-aarchxx.yml
+++ b/.github/workflows/ci-aarchxx.yml
@@ -47,6 +47,12 @@ jobs:
       with:
         submodules: true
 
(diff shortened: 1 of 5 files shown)


derekbruening:
32-bit whitelist_app (already on 64-bit ignore list): or should allow 2 potential errors, or get the test script to print out the false pos so we can figure it out.

PR #2348, opened 2020-12-13
i#2323 GA CI: Set up Linux package build

derekbruening:
First attempt to set up a package build.
Triggered via cron Fridays 9pm, or manually with a version input field.
Builds, creates a release with a tag, and posts the tarball.
Once we have the Linux x86 build working we can add the others.

Issue: #2323


derekbruening:
Going to merge so it will show up in Actions for easy testing.  This repo is low enough traffic I don't think it's worth forking just for these dev steps.

Issue #2361, opened 2021-01-16
uninit in NtContinue at init on GA CI


derekbruening:
This showed up in a mallocMD run once on GA CI:
https://github.com/DynamoRIO/drmemory/pull/2360/checks?check_run_id=1714548633
```
2021-01-16T19:19:52.8495990Z  48/118 Test  #48: mallocMD ..........................***Failed    4.67 sec

2021-01-16T19:19:52.8414391Z 48: CMake Error at runtest.cmake:572 (message):
2021-01-16T19:19:52.8415046Z 48:   stderr failed to match "~~Dr\.M~~ 2 unique, 2 total uninitialized
2021-01-16T19:19:52.8415793Z 48:   access\(es\)", found "~~Dr.M~~ 3 unique, 3 total uninitialized access(es)"
2021-01-16T19:19:52.8416349Z 48:   instead

2021-01-16T19:19:52.8348663Z 48: Error #1: UNINITIALIZED READ: reading register r10
2021-01-16T19:19:52.8349292Z 48: # 0 system call NtContinue parameter value #0
2021-01-16T19:19:52.8349817Z 48:     <system call>
2021-01-16T19:19:52.8350461Z 48: # 1 ntdll.dll!LdrInitializeThunk (0x00007ffa6c707cb8 <ntdll.dll+0x77cb8>)
2021-01-16T19:19:52.8351202Z 48:     ??:0
2021-01-16T19:19:52.8352400Z 48: Note: @0:00:00.228 in thread 1548
```

PR #2362, opened 2021-01-16
i#2361: Add NtContinue uninit to default suppressions

derekbruening:
Further analysis may show a programmatic solution, but for now we
auto-suppress to avoid tests failing on GA CI.

Issue: #2361


derekbruening:
Well mallocMD failed again: but this time over 0 vs 1 found leaks.  Grr, these tests are so flaky.  I think there's an umbrella issue on such a thing...looking.

PR #2366, opened 2021-01-19
i#2365: Cleanup: Remove non-USE_DRSYMS code and defines

### LINKED ISSUE #2365: [cleanup] remove !USE_DRSYMS define
Code for `!defined(USE_DRSYMS)` is very old and not supported now.  We should just remove the define.

derekbruening:
Removes the USE_DRSYMS define and all code only enabled when the
option is off, as we no longer support not using drsyms.

Removes the PERL_TO_EXE define as well as it was only still there for
the old post-process-symbolize model.

Fixes #2365


derekbruening:
failure is a flake: #2367

Issue #2367, opened 2021-01-19
GA CI tests failing due to missing or extra or different-sized leaks


derekbruening:
This is a container case for test failures involving leak counts or sizes.
Unfortunately, esp on Windows, we have test flakiness coming from leaks.

Here we have a 32-bit leak_indirect test which failed once:
https://github.com/DynamoRIO/drmemory/pull/2366/checks?check_run_id=1729970229
```
2021-01-19T19:04:18.6708707Z 5: CMake Error at runtest.cmake:534 (message):
2021-01-19T19:04:18.6709305Z 5:   stderr failed to match "~~Dr\.M~~ 2 unique, 2 total, 96 byte\(s\) of
2021-01-19T19:04:18.6709932Z 5:   leak\(s\)", found "~~Dr.M~~ 2 unique, 2 total, 80 byte(s) of leak(s)"
2021-01-19T19:04:18.6710408Z 5:   instead
```

derekbruening:
leak_indirect failed again in the same way: https://github.com/DynamoRIO/drmemory/pull/2369/checks?check_run_id=1757604049

derekbruening:
mallocMTd and mallocMD failed with:
https://github.com/DynamoRIO/drmemory/pull/2371/checks?check_run_id=1757756025
```
2021-01-24T17:03:47.4788307Z 55:   stderr failed to match "~~Dr\.M~~ 1 unique, 1 total, 59 byte\(s\) of
2021-01-24T17:03:47.4789135Z 55:   leak\(s\)", found "~~Dr.M~~ 0 unique, 0 total, 0 byte(s) of leak(s)"
2021-01-24T17:03:47.4789560Z 55:   instead
```

derekbruening:
I used tmate to connect to a GA CI VM that hits the failure.  I had to run several times to repro w/ logging:

```
replace_calloc 1 42
        carving out new chunk @0x017c3b20 => head=0x017c3b24, res=0x017c3b40
        replace_alloc_common arena=0x017b0000 flags=0x1004 request=42, align=8 alloc=48 => 0x017c3b40
# 0 replace_calloc                                    [d:\a\drmemory\drmemory\common\alloc_replace.c:2620]
# 1 main                                              [d:\a\drmemory\drmemory\tests\malloc.c:240]
get_shared_callstack: created pcs 0x1fe23da8
set range 0x017c3b40-0x017c3b6a => 0x0
        set byte 0x017c3b68
[...]

Issue #2341, opened 2020-12-11
umbra_client_faulty_redzone test hangs for 10 minutes on GA CI


derekbruening:
The umbra_client_faulty_redzone hangs, finally times out after 10 minutes (slowing down the CI), and then claims it passed:

```
2020-12-11T16:47:47.8429553Z         Start 136: umbra_client_faulty_redzone
2020-12-11T16:47:48.0482326Z 
2020-12-11T16:47:48.4508332Z 136: Test command: D:\a\drmemory\drmemory\build_drmemory-dbg-32\dynamorio\bin32\drrun.exe "-quiet" "-debug" "-client" "D:/a/drmemory/drmemory/build_drmemory-dbg-32/tests/umbra_client_faulty_redzone.client.dll" "0" "" "-msgbox_mask" "0" "-stderr_mask" "0xc" "--" "D:/a/drmemory/drmemory/build_drmemory-dbg-32/tests/umbra_app.exe"
2020-12-11T16:47:48.4637713Z 136: Test timeout computed to be: 600
2020-12-11T16:51:49.7198733Z 136: TEST PASSED
2020-12-11T16:57:47.8293867Z 145/145 Test #136: umbra_client_faulty_redzone .......***Timeout 600.01 sec
2020-12-11T16:57:47.8295647Z TEST PASSED
```

This is on Github Actions CI on Windows Server 2016.   I'm moving from Appveyor to there in #2334.  The log above is from https://github.com/DynamoRIO/drmemory/pull/2326/checks?check_run_id=1538421907

derekbruening:
I set a timeout of 60s (to shrink the CI time) and now that timeout lapsing is considered a failure: not sure what the difference is.

johnfxgalea:
Interesting, I will take a look. Does this happen only on Windows?

derekbruening:
> Interesting, I will take a look. Does this happen only on Windows?

I think so but it might be worth checking the Linux logs in case it was marked as passed and nobody noticed.

I just made the switch to GA for Windows and put this test on the runsuite_wrapper.pl ignore list to get the GA CI green: if you fix it please remove from the list.

derekbruening:
Well the test just failed on Linux: https://github.com/DynamoRIO/drmemory/runs/1541425935?check_suite_focus=true
It must be non-det there: it was not on the ignore list before and things were green.

derekbruening:
[...]

PR #2369, opened 2021-01-24
Update mangled operator names for x64

derekbruening:
The size_t type's mangled letter is m for 64-bit in the Itanium ABI
style.  We update the operator new mangled names for 64-bit to s/j/m/
here.  These are only used for distinguishing nothrow and placement,
not for locating operators, which may explain why there was no noticed
problem in the past.


derekbruening:
Failures are: umbra_client_faulty_redzone hang #2341, leak_indirect byte count mismatch #2367

PR #2373, opened 2021-01-31
Pull in static TLS DR bug fixes

derekbruening:
Updates DR to 39dc18f5 to pull in solutions to
DynamoRIO/dynamorio#4136 to fix crashes with static TLS use in
applications.


derekbruening:
32-bit Windows has 2 failures: mallocMD with a leak mismatch which is the flake #2367; and umbra_client_faulty_redzones, which crashed: never seen that before.  I filed it as #2375.  The prior run (I re-ran to it ensure #2374 is a flake) did not have the redzones test fail so I'm assuming the crash is unrelated to this DR update.

Issue #2375, opened 2021-01-31
CRASH umbra_client_faulty_redzone on GA CI 32-bit Windows


derekbruening:
This happened in the 2nd of 2 runs (re-ran):
https://github.com/DynamoRIO/drmemory/pull/2373/checks?check_run_id=1801981894
```
2021-01-31T18:17:22.8577386Z 136: Test command: D:\a\drmemory\drmemory\build_drmemory-dbg-32\dynamorio\bin32\drrun.exe "-quiet" "-debug" "-client" "D:/a/drmemory/drmemory/build_drmemory-dbg-32/tests/umbra_client_faulty_redzone.client.dll" "0" "" "-msgbox_mask" "0" "-stderr_mask" "0xc" "--" "D:/a/drmemory/drmemory/build_drmemory-dbg-32/tests/umbra_app.exe"
2021-01-31T18:17:22.8578904Z 136: Test timeout computed to be: 60
2021-01-31T18:17:23.1433153Z 136: <Exception in cache 0x1bf93047 interpreting DR code 0x7006e79a>
2021-01-31T18:17:23.1434336Z 136: <Application D:\a\drmemory\drmemory\build_drmemory-dbg-32\tests\umbra_app.exe (3200).  Tool internal crash at PC 0x7006e79a.  Please report this at your tool's issue tracker.  Program aborted.
2021-01-31T18:17:23.1435367Z 136: 0xc0000005 0x00badcad 0x7006e79a 0x7006e79a 0x00000001 0x1c049e75
2021-01-31T18:17:23.1435732Z 136: Base: 0x70050000
2021-01-31T18:17:23.1436157Z 136: Registers: eax=0x00000000 ebx=0x002fa000 ecx=0x00000000 edx=0x704be000
2021-01-31T18:17:23.1436659Z 136: 	esi=0x00000000 edi=0x00000000 esp=0x004ff9d4 ebp=0x004ff9d4
2021-01-31T18:17:23.1437050Z 136: 	eflags=0x00010216
2021-01-31T18:17:23.1437373Z 136: version 8.0.18658, custom build
[...]

PR #2376, opened 2021-01-31
i#2375: Fix faulty redzone test

johnfxgalea:
Fixes a bug in the test for faulty red zones. The bug led the test to potentially calculate app addresses with scratch registers not containing their original values.

Issue: #2375

### DIFF
--- a/tests/framework/umbra_client_faulty_redzone.c
+++ b/tests/framework/umbra_client_faulty_redzone.c
@@ -148,6 +148,7 @@ instrument_mem(void *drcontext, instrlist_t *ilist, instr_t *where,
         }
     }
 
+    // Found no memory operand to instrument.
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


johnfxgalea:
@derekbruening Please take a look. Hopefully, the issue is now resolved. I can't be certain as I didn't find the root cause of #2375 as I cannot reproduce locally.

johnfxgalea:
According to the CI, it seems like this PR didn't fix it. Marking the PR as draft, and will look again into it soon.

johnfxgalea:
We need to fix the issue via DynamoRIO (which I will do next).

However, this PR is still useful in making the test more reliable and correct.

@derekbruening Appreciate if you can take a look! Thank you.

derekbruening (approved):

johnfxgalea:
Created an issue for the new CI failure: https://github.com/DynamoRIO/drmemory/issues/2379

PR #2380, opened 2021-02-06
Fix doxygen 1.9.1 build error

derekbruening:
Removes the obsolete Doxyfile entry COLS_IN_ALPHA_INDEX.


derekbruening:
Windows failures are umbra_client_faulty_redzone #2375 and leak mismatches #2367.

PR #2381, opened 2021-02-06
DRi#4366: Rename list filter options

derekbruening:
Renames the following options:
- -lib_whitelist is now -lib_allowlist
- -lib_whitelist_frames is now -lib_allowlist_frames
- -src_whitelist is now -src_allowlist
- -src_whitelist_frames is now -src_allowlist_frames
- -lib_blacklist is now -lib_blocklist
- -lib_blacklist_frames is now -lib_blocklist_frames
- -check_uninit_blacklist is now -check_uninit_blocklist

The old option names still work for compatibility through a new
alternate-name field in the option macros, but they are not documented
(except in the old changelist).

Issue: DynamoRIO/dynamorio#4366


derekbruening:
Once again, Windows failures are umbra_client_faulty_redzone #2375 and leak mismatches #2367.

PR #2382, opened 2021-02-06
Fix drwrap crash by updating DR

derekbruening:
Updates DR to 795c8498 to bring in fix for
a drwrap crash which could affect Dr. Memory.

The DR update includes changes to asm_defines.asm,
which requires bumping Windows line numbers in some
test output patterns.

Issue: DynamoRIO/dynamorio#4607


derekbruening:
The procterm.nativeparent test looks like an actual DR regression in DynamoRIO/dynamorio#4703.  Investigating.

derekbruening:
The single failure is Windows 32-bit umbra_client_faulty_redzone which is #2375.

PR #2383, opened 2021-02-07
Clarify docs on blocklist and pause_at options

derekbruening:
For the docs on the blocklist, makes it more clear that all frames
must match.  Updates the actual default value.  Updates a few internal
references to the old names before PR #2381.

Adds instructions on how to switch to the app context for the
-pause_at options, using windbg.


derekbruening:
Once again, Windows failures are umbra_client_faulty_redzone #2375 and 32-bit Windows leak mismatches #2367.

PR #2384, opened 2021-02-07
i#2375,i#2367: Ignore/improve flaky umbra and leak tests

derekbruening:
Adds umbra_client_faulty_redzone to the ignore list until it is fixed
(#2375).

Relaxes the leak_indirect output to keep it greener (#2367).

Issue: #2375, #2367


derekbruening:
Woohoo, all green!

PR #2385, opened 2021-02-07
i#2080: Separate default and user-set blocklists

### LINKED ISSUE #2080: Add way to append to default lib_blacklist
The default lib_blacklist option is kind of complex, something like this depending on env vars:

`C:\Windows*.d??,C:\Program Files (x86)\Common Files\Microsoft Shared*.d??,C:\Program Files (x86)\Common Files\Microsoft Shared*.d??`

Now if a user wants to add one lib to this list, he has to duplicate the whole thing.  We should add a -lib_blacklist_append, or move the defaults to -lib_blacklist_default, or something.

derekbruening:
Moves the blocklist filled with default Windows paths to a new option
-lib_blocklist_default.  This lets the user add to the blocklist
without having to figure out what the default was and re-list it.

Updates the documentation.

Updates the existing blocklist test to test the new option behavior.
Relaxes the existing blocklist Windows test output so that it can be
removed from the ignore list.
Adds a new cross-platform test of the base blocklist feature.

Fixes #2080


derekbruening:
Windows failure is mallocMD not finding the leak it's supposed to: #2367.
OSX has 3 tests timeout which is weird. Re-running to see if it repros.

derekbruening:
The OSX hangs did not reproduce.  I filed #2386.

PR #2396, opened 2021-03-28
Add apt-get update step before every install

derekbruening:
For Github Actions, we've seen stale apt state in VM's.
We proactively run update before every install.


derekbruening:
Windows failure is leak mismatch: #2367

Issue #2260, opened 2020-02-16
Move drmemory.org to Github Pages, with automatic doxygen docs deployment from Travis


derekbruening:
This is the Dr. Memory equivalent of https://github.com/DynamoRIO/dynamorio/issues/2860 and  https://github.com/DynamoRIO/dynamorio/issues/3194.

derekbruening:
Success, after several tries:
https://travis-ci.com/DynamoRIO/drmemory/jobs/288271807
```
    Deploy DynamoRIO/drmemory_docs to github.com/DynamoRIO/drmemory_docs.git:master
```
And https://dynamorio.org/drmemory_docs/ says "2.3.18309"

PR #2397, opened 2021-03-28
i#2260 web: Produce doyxgen suitable for web embedding

derekbruening:
Copies DynamoRIO's method of producing doxygen suitable for
embedding in Github Pages jekyll.
Since we need custom menu tweaking, we now use our own
copy of CMake_rundoxygen rather than sharing DR's as-is.

Issue: #2260


derekbruening:
#2367 again.

PR #2400, opened 2021-04-05
DRi#4830: Downgrade to CMake 3.19.7 to avoid 32-bit failures

derekbruening:
CMake 3.20 on GA CI fails to link 32-bit binaries, so we downgrade
our jobs to 3.19.7 for now.

Issue: DynamoRIO/dynamorio#4830


derekbruening:
Win32 failure is leaks yet again: #2367

Issue #2016, opened 2017-08-07
Building natively on AArch64 fails


lgeek:
I'm running this on AArch64 system with Debian Stretch; Linux 4.9.0, gcc 6.3.0, cmake 3.7.2. It looks like the build system thinks it's running on x86-64. On the other hand, cmake prints `The system is: Linux - 4.9.0-3-arm64 - aarch64` in CMakeOutput.log. Here's the output from cmake:

```
$ cmake ..
-- The C compiler identification is GNU 6.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- The CXX compiler identification is GNU 6.3.0
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Dr. Memory version number: 1.11.17353
-- Looking for include file asm-i386/stat.h
-- Looking for include file asm-i386/stat.h - not found
-- Performing Test HAVE_FVISIBILITY_INTERNAL
-- Performing Test HAVE_FVISIBILITY_INTERNAL - Success
-- Performing Test HAVE_FVISIBILITY_HIDDEN
-- Performing Test HAVE_FVISIBILITY_HIDDEN - Success
-- Building DynamoRIO from local sources /home/cosmin/drmemory/build/dynamorio/cmake
-- Version number: 6.2.17316
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/as
/usr/bin/as missing flag "-mmnemonic.*intel"
/usr/bin/as too old: using /home/cosmin/drmemory/dynamorio/make/as-2.18.50
-- Found Perl: /usr/bin/perl (found version "5.24.1") 
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of uint
-- Check size of uint - done
-- Check size of ushort
-- Check size of ushort - done
-- Check size of bool
-- Check size of bool - failed
-- Check size of byte
-- Check size of byte - failed
-- Check size of sbyte
[...]

Issue #2379, opened 2021-02-02
CI test failure: hello


johnfxgalea:
We have a new CI failure concerning the hello test. It's pretty weird, as it concerns the log dir.

```
2021-02-02T17:46:43.2543900Z  1/13 Test  #3: hello ............................***Failed    0.16 sec
2021-02-02T17:46:43.2544750Z STDOUT: 
2021-02-02T17:46:43.2546240Z STDERR: ERROR: invalid -logdir: cannot find/write /Users/runner/work/drmemory/drmemory/build_drmemory-dbg-64/logs
2021-02-02T17:46:43.2547400Z 
2021-02-02T17:46:43.2547880Z 
2021-02-02T17:46:43.2548550Z CMake Error at runtest.cmake:316 (message):
2021-02-02T17:46:43.2549220Z   ***
2021-02-02T17:46:43.2551620Z   /Users/runner/work/drmemory/drmemory/build_drmemory-dbg-64/bin64/drmemory;-debug;-dr_debug;-dr;/Users/runner/work/drmemory/drmemory/build_drmemory-dbg-64/dynamorio;-batch;-dr_ops;-dumpcore_mask;-dr_ops;0;-callstack_style;0x27;-no_results_to_stderr;--;/Users/runner/work/drmemory/drmemory/build_drmemory-dbg-64/tests/hello
2021-02-02T17:46:43.2553740Z   has the wrong exit code (1): ERROR: invalid -logdir: cannot find/write
2021-02-02T17:46:43.2555110Z   /Users/runner/work/drmemory/drmemory/build_drmemory-dbg-64/logs
```

Ref: https://github.com/DynamoRIO/drmemory/pull/2376/checks?check_run_id=1816615166

PR #2404, opened 2021-04-28
i#2016: Port the umbra_64.c class to work on aarch64

gregcawthorne:
Enables shadow memory tracking on AArch64.

This is a subsection of the larger ongoing AArch64 port of DRMemory.

Issue: #2016

### DIFF
--- a/umbra/umbra_64.c
+++ b/umbra/umbra_64.c
@@ -1263,8 +1263,8 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
                                 int num_scratch_regs)
 {
 #if defined(AARCH64)
-    ASSERT(num_scratch_regs > umbra_num_scratch_regs_for_translation_arch(),
-        "umbra_insert_app_to_shadow_arch requires at least one scracth register provided");
+    if (num_scratch_regs >= umbra_num_scratch_regs_for_translation_arch())
+        return DRMF_ERROR_INVALID_PARAMETER;
 
     reg_id_t tmp = *scratch_regs;
 
--- a/umbra/umbra_64.c
+++ b/umbra/umbra_64.c
@@ -262,6 +262,7 @@ static dr_os_version_info_t os_version = {sizeof(os_version),};
 /* Each unit has 16 segments, which could be used for app or shadow. */
 #define NUM_SEGMENTS      16 /* 16 segments per unit */
 
+
 static ptr_uint_t seg_index_mask(uint num_seg_bits)
 {
     return (ptr_uint_t)(NUM_SEGMENTS - 1) << num_seg_bits;
@@ -1247,7 +1248,7 @@ umbra_value_in_shadow_memory_arch(IN    umbra_map_t *map,
 int
 umbra_num_scratch_regs_for_translation_arch()
 {
-    return 0;
+    return IF_AARCH64_ELSE(1, 0);
 }
 
 /* code sequence:
@@ -1263,6 +1264,9 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
                                 int num_scratch_regs)
 {
 #if defined(AARCH64)
+    ASSERT(num_scratch_regs > umbra_num_scratch_regs_for_translation_arch(),
+        "umbra_insert_app_to_shadow_arch requires at least one scracth register provided");
+
     reg_id_t tmp = *scratch_regs;
 
     instru_insert_mov_pc(drcontext, ilist, where, opnd_create_reg(tmp),
--- a/umbra/umbra_64.c
+++ b/umbra/umbra_64.c
@@ -1263,7 +1263,7 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
                                 int num_scratch_regs)
 {
 #if defined(AARCH64)
-    if (num_scratch_regs >= umbra_num_scratch_regs_for_translation_arch())
+    if (num_scratch_regs < umbra_num_scratch_regs_for_translation_arch())
         return DRMF_ERROR_INVALID_PARAMETER;
 
     reg_id_t tmp = *scratch_regs;
--- a/umbra/umbra_64.c
+++ b/umbra/umbra_64.c
@@ -263,16 +263,9 @@ static dr_os_version_info_t os_version = {sizeof(os_version),};
 #define NUM_SEGMENTS      16 /* 16 segments per unit */
 
 
-#ifdef AARCH64
-// This simply masks off the lower 2 bits which must be 0 for
-// a valid address.
-#define AARCH64_SEG_MASK 0xfffffffffffffffc
-#endif
-
-
 static ptr_uint_t seg_index_mask(uint num_seg_bits)
 {
-    return IF_AARCH64_ELSE(0x0, (ptr_uint_t)(NUM_SEGMENTS - 1) << num_seg_bits);
+    return (ptr_uint_t)(NUM_SEGMENTS - 1) << num_seg_bits;
 }
 
 static ptr_uint_t segment_size(uint num_seg_bits)
@@ -282,7 +275,7 @@ static ptr_uint_t segment_size(uint num_seg_bits)
 
 static ptr_uint_t segment_mask(uint num_seg_bits)
 {
-    return IF_AARCH64_ELSE(AARCH64_SEG_MASK, segment_size(num_seg_bits) - 1);
+    return segment_size(num_seg_bits) - 1;
 }
 
 static ptr_uint_t segment_base(uint num_seg_bits, app_pc pc)
@@ -363,18 +356,6 @@ static ptr_uint_t map_disp_win81[] = {
  * We check conflicts later when creating shadow memory.
  */
 #ifdef LINUX
-#ifdef AARCH64
-static const app_segment_t app_segments_initial[] = {
-    {(app_pc)0x0000ff0000000000,  (app_pc)0x0001000000000000, 0},
-    {(app_pc)0x0000000000000000,  (app_pc)0x0000010000000000, 0},
-    { NULL, NULL, 0 },
-    { NULL, NULL, 0 },
-    { NULL, NULL, 0 },
-    { NULL, NULL, 0 },
-    { NULL, NULL, 0 },
-};
-
-#else
 static const app_segment_t app_segments_initial[] = {
     /* We split app3 [0x7F0000000000, 0x800000000000) into two parts:
      * [0x7F0000000000, 0x7FFFFF400000) and [0x7FFFFF800000, 0x800000000000).
@@ -408,7 +389,6 @@ static const app_segment_t app_segments_initial[] = {
     { NULL, NULL, 0 },
     { NULL, NULL, 0 },
 };
-#endif
 static const app_segment_t app_segments_initial_no_vsyscall[] = {
     /* We do not need any gaps or splitting without vsyscall. */
     /* for all additional segments */
@@ -1285,8 +1265,16 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
 {
 #if defined(AARCH64)
     reg_id_t tmp = *scratch_regs;
+
+    instru_insert_mov_pc(drcontext, ilist, where, opnd_create_reg(tmp),
+                             OPND_CREATE_INT64(map->mask));
+    PRE(ilist, where, INSTR_CREATE_and(drcontext,
+                                       opnd_create_reg(reg_addr), opnd_create_reg(reg_addr),
+                                       opnd_create_reg(tmp)));
+
     instru_insert_mov_pc(drcontext, ilist, where, opnd_create_reg(tmp),
                                 OPND_CREATE_INT64(map->disp));
+
     PRE(ilist, where, INSTR_CREATE_add(drcontext,
                                        opnd_create_reg(reg_addr), opnd_create_reg(reg_addr),
                                        opnd_create_reg(tmp)));
@@ -1299,11 +1287,6 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
                                            opnd_create_reg(reg_addr),
                                            OPND_CREATE_INT8(map->shift)));
     }
-    instru_insert_mov_pc(drcontext, ilist, where, opnd_create_reg(tmp),
-                             OPND_CREATE_INT64(map->mask));
-    PRE(ilist, where, INSTR_CREATE_and(drcontext,
-                                       opnd_create_reg(reg_addr), opnd_create_reg(reg_addr),
-                                       opnd_create_reg(tmp)));
 #else
     PRE(ilist, where, INSTR_CREATE_and(drcontext,
                                        opnd_create_reg(reg_addr),
--- a/umbra/umbra_64.c
+++ b/umbra/umbra_64.c
@@ -26,6 +26,7 @@
  */
 
 #include "dr_api.h"
+#include "instru.h"
 #include "umbra.h"
 #include "umbra_private.h"
 #include "drmemory_framework.h"
@@ -261,10 +262,13 @@ static dr_os_version_info_t os_version = {sizeof(os_version),};
 /* Each unit has 16 segments, which could be used for app or shadow. */
 #define NUM_SEGMENTS      16 /* 16 segments per unit */
 
+
+#ifndef AARCH64
 static ptr_uint_t seg_index_mask(uint num_seg_bits)
 {
     return (ptr_uint_t)(NUM_SEGMENTS - 1) << num_seg_bits;
 }
+#endif
 
 static ptr_uint_t segment_size(uint num_seg_bits)
 {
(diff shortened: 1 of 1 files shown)


umbra/umbra_64.c
```diff
@@ -354,6 +358,37 @@ static ptr_uint_t map_disp_win81[] = {
  * We check conflicts later when creating shadow memory.
  */
 #ifdef LINUX
+#ifdef AARCH64
+static const app_segment_t app_segments_initial[] = {
+    /* We split app3 [0x7F0000000000, 0x800000000000) into two parts:
```
johnfxgalea, line 363: I see that a big chunk of this comment block was copied from that below. Consider condensing to avoid duplicate text, as future updates would then require double changes.

umbra/umbra_64.c
```diff
@@ -839,7 +875,7 @@ umbra_map_arch_init(umbra_map_t *map, umbra_map_options_t *ops)
     ASSERT(map->shadow_block_size >= ALLOC_UNIT_SIZE &&
            map->app_block_size    >= ALLOC_UNIT_SIZE,
            "block size too small");
-    map->mask = segment_mask(num_seg_bits) | seg_index_mask(num_seg_bits);
+    map->mask = IF_AARCH64_ELSE(0xfffffffffffffffc, segment_mask(num_seg_bits) | seg_index_mask(num_seg_bits));
```
johnfxgalea, line 878: - Avoid magic values in code as this hinders readability. 
- Is it possible to have segment_mask() return the mask, rather than have the IF_AARCH64_ELSE  macro here?
- Finally, add a comment on how this mask was derived.

johnfxgalea (changes requested):
Thank you for the PR! Nice to see support for AARCH64.

I think it is best that we have some tests just to ensure all is working, and avoid regression.

gregcawthorne:
> Thank you for the PR! Nice to see support for AARCH64.
> 
> I think it is best that we have some tests just to ensure all is working, and avoid regression.

I was hoping the fact that the existing umbra tests pass with the full port, might suffice.

But we could add more specific tests.

johnfxgalea:
@derekbruening Does Dr. Memory's CI run tests for AArch64? IIRC, they used to run prior to the switch to Github Actions?

gregcawthorne:
> @derekbruening Does Dr. Memory's CI run tests for AArch64? IIRC, they used to run prior to the switch to Github Actions?

If I recall correctly prior to this body of work DRMemory didn't even compile on AArch64. It only had 32 bit arm support. However this might have changed since I started working on it, or perhaps I misunderstood.

johnfxgalea:
> > @derekbruening Does Dr. Memory's CI run tests for AArch64? IIRC, they used to run prior to the switch to Github Actions?
> 
> If I recall correctly prior to this body of work DRMemory didn't even compile on AArch64. It only had 32 bit arm support. However this might have changed since I started working on it, or perhaps I misunderstood.

Sorry, I meant 32-bit ARM

derekbruening:
> > > @derekbruening Does Dr. Memory's CI run tests for AArch64? IIRC, they used to run prior to the switch to Github Actions?
> > 
> > 
> > If I recall correctly prior to this body of work DRMemory didn't even compile on AArch64. It only had 32 bit arm support. However this might have changed since I started working on it, or perhaps I misunderstood.
> 
> Sorry, I meant 32-bit ARM

On Travis there was a 32-bit ARM cross-compile job, but no execution.

What should be done is to add both a32 and a64 QEMU runs (at least some simple tests are likely to run on QEMU given that they do for plain DR) and maybe reuse the DR Jenkins machine and setup to run a64 tests as support is added.

johnfxgalea:
> > > > @derekbruening Does Dr. Memory's CI run tests for AArch64? IIRC, they used to run prior to the switch to Github Actions?
> > > 
> > > 
> > > If I recall correctly prior to this body of work DRMemory didn't even compile on AArch64. It only had 32 bit arm support. However this might have changed since I started working on it, or perhaps I misunderstood.
> > 
> > 
> > Sorry, I meant 32-bit ARM
> 
> On Travis there was a 32-bit ARM cross-compile job, but no execution.
> 
> What should be done is to add both a32 and a64 QEMU runs (at least some simple tests are likely to run on QEMU given that they do for plain DR) and maybe reuse the DR Jenkins machine and setup to run a64 tests as support is added.

I see... Okay, so in that case, unfortunately we will punt test for this PR.

@derekbruening Is there an issue page regarding the above? I couldn't find it.

johnfxgalea:
@gregcawthorne No rush, just let me know when ready for me to take another look at the PR.

derekbruening:
> @derekbruening Is there an issue page regarding the above? I couldn't find it.

Probably not yet in the DrM tracker: the DR QEMU issue is https://github.com/DynamoRIO/dynamorio/issues/4719

umbra/umbra_64.c
gregcawthorne: I have now made these changes.

umbra/umbra_64.c
gregcawthorne: Removed comment.

gregcawthorne:
> @gregcawthorne No rush, just let me know when ready for me to take another look at the PR.

I have made the changes you have suggested, but haven't added a new test.

I am currently unsure of the ci-windows/vs2017-32 build failure in CI.

umbra/umbra_64.c
```diff
@@ -354,6 +363,18 @@ static ptr_uint_t map_disp_win81[] = {
  * We check conflicts later when creating shadow memory.
  */
 #ifdef LINUX
+#ifdef AARCH64
+static const app_segment_t app_segments_initial[] = {
+    {(app_pc)0x0000ff0000000000,  (app_pc)0x0001000000000000, 0},
```
johnfxgalea, line 368: Just having another look, and I don't understand why the layout needs to be different from the existing one. Can you explain, or provide a reference, regarding how you defined this app segment layout please?

umbra/umbra_64.c
```diff
@@ -1261,6 +1283,28 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
                                 reg_id_t *scratch_regs,
                                 int num_scratch_regs)
 {
+#if defined(AARCH64)
+    reg_id_t tmp = *scratch_regs;
+    instru_insert_mov_pc(drcontext, ilist, where, opnd_create_reg(tmp),
+                                OPND_CREATE_INT64(map->disp));
+    PRE(ilist, where, INSTR_CREATE_add(drcontext,
```
johnfxgalea, line 1290: The instruction sequence is different from the existing one. One did you place the `and` instruction last here?

johnfxgalea:
> > @gregcawthorne No rush, just let me know when ready for me to take another look at the PR.
> 
> I have made the changes you have suggested, but haven't added a new test.
> 
> I am currently unsure of the ci-windows/vs2017-32 build failure in CI.

@gregcawthorne Thank you for the changes. Yes, we will skip tests due to limitations of the CI, but I am assuming you tried the code locally, and all works fine, right?

johnfxgalea:
@gregcawthorne Ignore the failure on the Windows machine. It failed due to a known flaky test (ref: https://github.com/DynamoRIO/drmemory/issues/2361)

umbra/umbra_64.c
gregcawthorne: Okay I will review this section and get back to you.

umbra/umbra_64.c
```diff
@@ -1261,6 +1283,28 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
...
+                                           opnd_create_reg(reg_addr), opnd_create_reg(reg_addr),
+                                           OPND_CREATE_INT8(map->shift)));
+    } else if (map->options.scale <= UMBRA_MAP_SCALE_DOWN_2X) {
+        PRE(ilist, where, XINST_CREATE_slr_s(drcontext,
+                                           opnd_create_reg(reg_addr),
+                                           OPND_CREATE_INT8(map->shift)));
+    }
+    instru_insert_mov_pc(drcontext, ilist, where, opnd_create_reg(tmp),
+                             OPND_CREATE_INT64(map->mask));
+    PRE(ilist, where, INSTR_CREATE_and(drcontext,
```
johnfxgalea, line 1304: Why did you place the `and` instruction last here, in contrast to the existing code below?

gregcawthorne: yes I have done:
add => shift => and

rather than the existing:
and => add => shift

My understanding is that the purpose is the shift it into some other distinct space range with the correct scale and the order of operations didn't matter too much as long as 'a' mapping was achieved.

AArch64 doesn't use the same segmentation scheme as x86 as far as I know.

Therefore I am using the segment mask variable to simply mask off the lower 2 bits of the address in order to secure a valid address (as commented), and this makes sense to be done as the final operation.

gregcawthorne:
Hi guys. Just giving this a nudge.

johnfxgalea:
> Hi guys. Just giving this a nudge.

> Okay I will review this section and get back to you.

Hey thanks for the nudge.

I was under the impression that I am still awaiting my above comments to be addressed. Particularly: https://github.com/DynamoRIO/drmemory/pull/2404#discussion_r623011141

gregcawthorne:
> > Hi guys. Just giving this a nudge.
> 
> > Okay I will review this section and get back to you.
> 
> Hey thanks for the nudge.
> 
> I was under the impression that I am still awaiting my above comments to be addressed. Particularly: [#2404 (comment)](https://github.com/DynamoRIO/drmemory/pull/2404#discussion_r623011141)

Yes I will address that. I was thinking too sequentially.

umbra/umbra_64.c
johnfxgalea: > yes I have done:
> add => shift => and
> 
> rather than the existing:
> and => add => shift
> 
> My understanding is that the purpose of the mapping is to shift and scale into some other distinct space range with the correct scale. The order of operations doesn't matter too much as long as 'a' mapping was achieved.
> 

I think it can make a difference. The `and` is used to mask the address, the `add` is used as a displacement value to reach a shadow region, and the `shift` is intended for scaling. While it may not make a difference in some cases, having the `and` last might clear information injected earlier by the displacement/scaling if we make future changes.

In any case, it is much better to have the same code sequence as other platforms if possible, as otherwise it appears we are doing something special and potentially arises confusion. 

> AArch64 doesn't use the same segmentation scheme as x86 as far as I know.
> 
> Therefore I am using the mask variable to simply mask off the lower 2 bits of the address in order to secure a valid address (as commented), and this makes sense to be done as the final operation.

The mask and displacement values are critically set up in order to have shadow memory regions not collide with app regions. These values were carefully crafted, and some recently generated with the help of automated tools (https://github.com/DynamoRIO/drmemory/blob/master/tools/umbra_layout.py).

Maybe I am missing something, but I still don't understand why we can't use the existing segment mapping for Linux?

johnfxgalea:
> > > Hi guys. Just giving this a nudge.
> > 
> > 
> > > Okay I will review this section and get back to you.
> > 
> > 
> > Hey thanks for the nudge.
> > I was under the impression that I am still awaiting my above comments to be addressed. Particularly: [#2404 (comment)](https://github.com/DynamoRIO/drmemory/pull/2404#discussion_r623011141)
> 
> Yes I will address that. I was thinking too sequentially.

Sure, no problem! Thank you.

I added an other comment (https://github.com/DynamoRIO/drmemory/pull/2404#discussion_r628136173) relating to the above.

umbra/umbra_64.c
gregcawthorne: Okay so I when you start up DRMemory does a scan of all the current memory blocks. I am pretty sure  I simply looked at the default ones I was getting every time on my taishan AArch64 system and copied them in.

gregcawthorne:
Hi.

Sorry for my lateness on this, was out for a week and also missed the notifications.

I took onboard what you have said and have reverted the behaviour you were questioning (different initial segment map + incorrect order of the and/add/shift) and re ran the tests we have ported and we are getting the same tests passing/failing.

I will chalk this up to my early brash guess work when I first started orienting myself.

Thank you!

gregcawthorne:
I can't see currently how my changes could necessarily break x86 on only osx.

johnfxgalea:
@gregcawthorne You're right. The failure doesn't concern your code; it concerns this issue: https://github.com/DynamoRIO/drmemory/issues/2379

umbra/umbra_64.c
```diff
@@ -1261,6 +1262,31 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
                                 reg_id_t *scratch_regs,
                                 int num_scratch_regs)
 {
+#if defined(AARCH64)
+    reg_id_t tmp = *scratch_regs;
```
johnfxgalea, line 1266: You are relying on the user to reserve a scratch register; hence the dereference. Please add a check to ensure that the num_scratch_regs conforms to this assumption.

Also, fix umbra_num_scratch_regs_for_translation_arch() so that the user is aware that we require a scratch register.

johnfxgalea (commented):
I had another look. Seems like we are progressing well.

Thank you!

umbra/umbra_64.c
gregcawthorne: Okay thanks! Have just pushed my interpretation of what you have asked for.

johnfxgalea: Sorry, I was not clear. Please have DRMF_ERROR_INVALID_PARAMETER be returned, instead of using an ASSERT.

Moreover, I believe the check needs to be >= and not >.

umbra/umbra_64.c
```diff
@@ -1261,6 +1262,34 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
...
                                 int num_scratch_regs)
 {
+#if defined(AARCH64)
+    ASSERT(num_scratch_regs > umbra_num_scratch_regs_for_translation_arch(),
+        "umbra_insert_app_to_shadow_arch requires at least one scracth register provided");
+
+    reg_id_t tmp = *scratch_regs;
+
+    instru_insert_mov_pc(drcontext, ilist, where, opnd_create_reg(tmp),
+                             OPND_CREATE_INT64(map->mask));
```
johnfxgalea, line 1272: Please fix indentation; the second line needs to be under drcontext.

gregcawthorne: done

umbra/umbra_64.c
```diff
@@ -1261,6 +1262,34 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
...
+    PRE(ilist, where, INSTR_CREATE_and(drcontext,
+                                       opnd_create_reg(reg_addr), opnd_create_reg(reg_addr),
+                                       opnd_create_reg(tmp)));
+
+    instru_insert_mov_pc(drcontext, ilist, where, opnd_create_reg(tmp),
```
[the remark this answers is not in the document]
gregcawthorne: done

umbra/umbra_64.c
gregcawthorne: ah okay yes, done

umbra/umbra_64.c
```diff
@@ -1261,6 +1262,34 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
...
+    PRE(ilist, where, INSTR_CREATE_add(drcontext,
+                                       opnd_create_reg(reg_addr), opnd_create_reg(reg_addr),
+                                       opnd_create_reg(tmp)));
+    if (map->options.scale == UMBRA_MAP_SCALE_UP_2X) {
+        PRE(ilist, where, INSTR_CREATE_lsl(drcontext,
+                                           opnd_create_reg(reg_addr), opnd_create_reg(reg_addr),
+                                           OPND_CREATE_INT8(map->shift)));
+    } else if (map->options.scale <= UMBRA_MAP_SCALE_DOWN_2X) {
+        PRE(ilist, where, XINST_CREATE_slr_s(drcontext,
+                                           opnd_create_reg(reg_addr),
```
johnfxgalea, line 1289: Fix Indentation; opnd_create_reg(reg_addr), needs to be under drcontext.

umbra/umbra_64.c
```diff
@@ -1261,6 +1262,34 @@ umbra_insert_app_to_shadow_arch(void *drcontext,
                                 reg_id_t *scratch_regs,
                                 int num_scratch_regs)
 {
+#if defined(AARCH64)
+    if (num_scratch_regs >= umbra_num_scratch_regs_for_translation_arch())
```
johnfxgalea, line 1266: This is now incorrect :) We need to return DRMF_ERROR_INVALID_PARAMETER if num_scratch_regs is less than the required number of scratch registers as determined via umbra_num_scratch_regs_for_translation_arch().

Recall `reg_id_t *scratch_regs` and `int num_scratch_regs` are user controlled parameters. For umbra_insert_app_to_shadow_arch() to work, the user needs to provide a sufficient number of scratch registers. It is for this reason why we are including the check.

johnfxgalea (changes requested):
I think these are the final comments and once they are fixed, we will be very close to merging. Thank you.

umbra/umbra_64.c
gregcawthorne: yep, wasn't really thinking evidently, but woke up this morning and corrected (before reading this :P)

johnfxgalea: Awesome! Thanks for the quick fix. We just have the last pending issue regarding indentation, and then we are good for merging.

umbra/umbra_64.c
gregcawthorne: done

gregcawthorne:
Okay, do you guys have any preferences of which part of the larger initial aarch64 port would be 'best' to upstream next?

johnfxgalea (approved):
LGTM. Thank you!

johnfxgalea:
> Okay, do you guys have any preferences of which part of the larger initial aarch64 port would be 'best' to upstream next?

Not really, just try to keep the PRs as small as possible to facilitate reviewing!

It would be great if we consider adding the testing architecture for aarch64 before adding more code.(See: https://github.com/DynamoRIO/drmemory/pull/2404#issuecomment-828676715). If you can help with this, it would be really appreciated (but I understand if you are busy).

johnfxgalea:
@gregcawthorne Merged! Thanks again for the PR!

gregcawthorne:
Hi John, I can add any chances I have made to the tests next sure.

The port actually depends on dynamorio patch I have which is to add the ability to use memory references as arguments in clean calls.

I will work to get that upstreamed too.

johnfxgalea:
> Hi John, I can add any changes I have made to the tests next sure.
> 
> The port actually depends on dynamorio patch I have which is to add the ability to use memory references as arguments in clean calls.
> 
> I will work to get that upstreamed too.

Okay, no problem. Feel free to request me as a reviewer for the DR patch if you wish.

PR #2410, opened 2021-06-30
i2016 aarch64 drsyscall port

gregcawthorne:
Porting the drsyscall sub tool to AArch64 which is from the larger DrMemory port branch.


derekbruening:
(Please request a review from me or @johnfxgalea when ready for review)

(Windows failures may be #2367 which are hindering development -- if anyone has cycles to look at them or maybe just add to ignore list all tests we've seen fail (unfortunately includes some key ones) to get it green that would be awesome)

derekbruening (approved):

Issue #2417, opened 2021-09-06
Upgrade GA CI from Ubuntu 16.04 which is no longer supported


derekbruening:
We need to upgrade from 16.04:

```
The ubuntu-16.04 environment is deprecated and will be removed on September 20, 2021. Migrate to ubuntu-latest instead. For more details see https://github.com/actions/virtual-environments/issues/32
```

Xref https://github.com/DynamoRIO/dynamorio/issues/4953

PR #2423, opened 2021-09-26
i#2417 ubuntu20: Fix new libc unaddr and blocklist error

derekbruening:
Fixes two failing tests on Ubuntu18+:

Adds a new string unaddr pattern where recent libc
reads a whole word beyond the allocated bounds.
This fixes a false positive in the loader test
on Ubuntu18+.

Fixes an uninitialized field tls_report_t.last_query_mod_start which
was failing the blocklist_uninit.op test.

Issue: #2417


derekbruening:
Windows failures are the !@$(@*%$ doxygen sourceforge.

PR #2463, opened 2022-09-24
Support gcc 12

derekbruening:
Fixes build warnings with gcc 12.
Updates DR to 44312ad1c to fix gcc 12 issues there.


derekbruening:
Mac is hello test with a reported mem leak.  Seems to be non-det.  Will separate it out: likely triggered by macos-11 update here.

PR #2490, opened 2024-01-13
i#2485: Update DR to use elfutils for DWARF-5 support

### LINKED ISSUE #2485: DWARF5 is not supported by drsyms, resulting in no line numbers with gcc 11+
Building Dr. Memory on a system with gcc 12.2, the tests fail b/c they can't get line numbers:

```
Error #1: UNADDRESSABLE ACCESS beyond heap bounds: reading 0x00007f796fc01223-0x00007f796fc01224 1 byte(s)
# 0 main   (0x00007f796fbfb33a <malloc+0x133a>)        
    ??:0                                                                                  
```

It's because the test binary has DWARF5:
```
$ readelf --debug-dump=info tests/malloc | grep -A 2 'Compilation Unit @'
  Compilation Unit @ offset 0:
   Length:        0x78d (32-bit)
   Version:       5
```

Dr. Memory uses DynamoRIO's drsyms library, and it uses elftoolchain, picked due to its BSD instead of GPL license.  However, it is not kept up to date these days and lacks DWARF5 support: https://sourceforge.net/p/elftoolchain/tickets/611.

This is filed in the DynamoRIO tracker as DynamoRIO/dynamorio#5926.

Adding -gdwarf-4: the test works.

derekbruening:
Updates DR to 3e3b976e316e9469ca36338fb7f3fa086739d03a which pulls in the use of elfutils in drsyms for DWARF-5 support on Linux.

Updates the documentation and flags for Linux to remove -gdwarf-4.

Fixes #2485

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # **********************************************************
-# Copyright (c) 2010-2023 Google, Inc.  All rights reserved.
+# Copyright (c) 2010-2024 Google, Inc.  All rights reserved.
 # Copyright (c) 2009-2010 VMware, Inc.  All rights reserved.
 # **********************************************************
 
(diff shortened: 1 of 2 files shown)


derekbruening:
The test failures are 32-bit AMD: filed #2491.  WIll try to address separately.

Issue #2491, opened 2024-01-14
ASSERT "vsyscall incorrect assumption" on AMD 32-bit


derekbruening:
This is now showing up on the GA CI VM's since they now use AMD hardware.

Xref DynamoRIO/dynamorio#6417 where 32-bit AMD machines added to GA CI caused a lot of test
failures.  This looks similar.

https://github.com/DynamoRIO/drmemory/actions/runs/7515048417/job/20458582246?pr=2490
```
15:   ~~Dr.M~~ ASSERT FAILURE (thread 43342):
15:   /home/runner/work/drmemory/drmemory/drsyscall/drsyscall_linux.c:239:
15:   !is_using_sysint() (vsyscall incorrect assumption)
```

PR #2495, opened 2024-02-07
DRi#6611: Fix glibc 2.37 elfutils hang

derekbruening:
Updates DR to b42b82b1d to fix a problem with DR's private loader with glibc 2.37 and the new elfutils libdw when addresses are looked up.

Issue: DynamoRIO/dynamorio#6611


derekbruening:
The failure are 32-bit AMD #2491

PR #2497, opened 2024-03-18
Update WIX version to 6.14

derekbruening:
Updates the PATH of our package build to find the new WIX version on the GA CI VM's.

Issue: DynamoRIO/dynamorio#6709


derekbruening:
The Linux failures are AMD vsyscall asserts: #2491

PR #2510, opened 2024-09-13
Suggest using -gdwarf-2 instead of -ggdb with mingw64.

brettle:
At least with mingw-w64-x86_64-gcc 14.2.0, line numbers are not provided when using -ggdb but are when using -gdwarf-2. Fix was found here: https://stackoverflow.com/questions/73508016/dr-memory-does-not-show-line-number-in-call-stack


derekbruening (approved):
Thank you for the patch.

Issue #2498, opened 2024-04-25
Drmemory.exe unable yto load client library:bcrpt.dll: library initializer failed


chenyongquan2:
**Describe the bug**
When I try to use Drmemory.exe to analyse my exe to find problem, it shows the following error:
Unbale to load client library: bcrypt.dll: library initializer failed

![image](https://github.com/DynamoRIO/drmemory/assets/101924391/6c1716ba-f1fd-4611-80d2-9d2619602cfc)

But I can found the bcrypt.dll in n my system file
![image](https://github.com/DynamoRIO/drmemory/assets/101924391/bc6cf923-532e-4661-9e1a-0750ecb65227)

![image](https://github.com/DynamoRIO/drmemory/assets/101924391/d858686b-d0f3-4a9a-bbb4-a01f9eb06628)

And my version info:
![image](https://github.com/DynamoRIO/drmemory/assets/101924391/1d1fae51-ed04-460f-aea1-2906584cbb16)

How could I fix these problem

Hzlin7:
![image](https://github.com/DynamoRIO/drmemory/assets/31055209/6c5d0df8-add3-4b68-850a-9b58411571de)
I got the same problem here，have you solved the problem yet?

derekbruening:
I think this was fixed by DynamoRIO/dynamorio#6571

derekbruening:
Pulled into drmemory by #2495

PR #2515, opened 2024-09-23
i#2498 init failures: Ignore privlib init failures

derekbruening:
Updates DR to 51b152fa2 extends the private loader bcrypt.dll entry failure being ignored to ignore all private library entry failures. We have had multiple cases of initializers failing without fatal consequences, and given that we do not have resources to track down every detail it is best to note the failure and move on. We have seen Dr. Memory in particular fail on newer Windows versions and in some cases ignoring that failure lets the tool continue and succeed, as only a small piece of the private library is needed (generally pulled in as a chain of dependences from dbghelp.dll which is used for symbol information). The private loader on Windows is considered best-effort at this point.

Issue: #2498


derekbruening:
The failures are the vsyscall ones from Github switching to AMD: #2491

Issue #2342, opened 2020-12-11
many 32-bit tests hit DR assert core\ir\x86\encode.c:2722 CHECK_TRUNCATE_TYPE_sbyte(new_offs) on GA CI


derekbruening:
```
2020-12-11T16:42:05.5500012Z <Application D:\a\drmemory\drmemory\build_drmemory-dbg-32\tests\malloc.exe (820).  Internal Error: DynamoRIO debug check failure: ..\dynamorio\core\ir\x86\encode.c:2722 CHECK_TRUNCATE_TYPE_sbyte(new_offs)
```
Looks like the following tests hit this on GA CI:
malloc, cs2bug, operators, wincrt, wincrtdbg, cs2bugMTd, operatorsMDd, leak_string.

Is it a short jmp in DrM instru that doesn't reach??  I thought that hit a
different error though during encoding?

derekbruening:
Actually it is these tests (only looked at executable name before):
+ wrap_malloc
+ wrap_cs2bug
+ wrap_operators
+ wrap_wincrt
+ wrap_wincrtdbg
+ wrap_cs2bugMTd
+ wrap_operatorsMDd
+ leak_string

So it's the wrap_ tests: leak_string also sets -no_replace_malloc.

derekbruening:
I was thinking of dropping wrap_ support anyway.  Not enough developers to support extra options.

Issue #2202, opened 2019-06-25
ASSERT pcache-use "meta-instr faulted?"


derekbruening:
Started happening in a PR but unlikely to be related:

https://travis-ci.com/DynamoRIO/drmemory/jobs/210644641

```
<Application /home/travis/build/DynamoRIO/drmemory/build_drmemory-dbg-32/tests/pcache (419) DynamoRIO usage error : meta-instr faulted?  must set translation field and handle fault!>
<Usage error: meta-instr faulted?  must set translation field and handle fault! (/home/travis/build/DynamoRIO/drmemory/dynamorio/core/translate.c, line 837)
```

derekbruening:
Xref https://github.com/DynamoRIO/dynamorio/issues/1369

PR #2526, opened 2024-10-04
i#2522 strcasestr prototype error: Remove `const` from result type

### LINKED ISSUE #2522: BUILD: drmemory build failure caused by strcasestr declaration error
**Describe the bug**

strcasestr's definition is char *(const char *, const char *) whereas drmemory/common/utils.h uses the older
const char *(const char *, const char *).
This causes a build failure on linux (64-bit, recent vintage).
```
In file included from /tmp/drmemory/common/utils_shared.c:30:
/usr/include/string.h:380:14: error: conflicting types for ‘strcasestr’; have ‘char *(const char *, const char *)’
  380 | extern char *strcasestr (const char *__haystack, const char *__needle)
      |              ^~~~~~~~~~
In file included from /tmp/drmemory/common/utils_shared.c:28:
/tmp/drmemory/common/utils.h:1015:1: note: previous declaration of ‘strcasestr’ with type ‘\
const char *(const char *, const char *)’
 1015 | strcasestr(const char *text, const char *pattern);
      | ^~~~~~~~~~
```

**To Reproduce**

$ git clone https://github.com/DynamoRIO/drmemory.git
$ cd drmemory
$ ./make/git/dev-setup.sh
$ cd ..
$ mkdir build
$ cd build
$ cmake ../drmemory
$ make -k # other failures may be present: -k keeps going so that strcasestr failure is seen

**Versions**
 - What version of Dr. Memory are you using? git head @bfbc4a118a58a182770d306120cad41b9893c53

xdje42:
The definition of strcasestr is

char *strcasestr(const char *text, const char *pattern);

but drmemory uses

const char *strcasestr(const char *text, const char *pattern);

This causes build errors when both /usr/include/string.h and common/utils.h are included in linux. The definition for strcasestr for C has always not had the `const` in the result type (going back to its original addition to glibc in 1997). Thus we're pretty safe in not breaking anything on *nix. The other main use case of drmemory's strcasestr is on Windows, which does not have strcasestr.

Tested:
$ cmake && make
Also, there are failures in the clang and x86 workflows, but comparing the logs before/after
shows all failures pre-exist this patch, no new failures.

Here is the complete audit, all failures are pre-existing.

Recorded (errantly) as i#1938 failures:
  - wrap_operators: "internal crash" (Received SIGSEGV)
    - filed as #2535
  - wrap_cs2bug: CHECK_TRUNCATE_TYPE_sbyte(new_offs) (#2342)
  - pcache-use: "Usage error: meta-instr faulted? must set translation field and handle fault!" (#2202)
  - app_suite: vsyscall incorrect assumption (#2491)
  - app_suite.pattern: assert failure: save->ignore_next_delete == 0 (premature deletion)
    - filed as i#2537
  - fuzz_threads: vsyscall incorrect assumption (#2491)

Note: The above are marked as expected failures but [runsuite_wrapper.pl hardwires the addition of "i#1938" to test output](https://github.com/DynamoRIO/drmemory/blob/master/tests/runsuite_wrapper.pl#L281). #2534

Not currently recorded as expected failures (all #2491 - vsyscall incorrect assumption):
  - selfmod
  - syscalls_unix
  - clone
  - pthread_test
  - realloc

Fixes #2522

### DIFF
--- a/common/utils.h
+++ b/common/utils.h
@@ -1011,7 +1011,7 @@ drmem_strndup(const char *src, size_t max, heapstat_t type);
 #define MAX_OPTION_LEN DR_MAX_OPTIONS_LENGTH
 
 #if !defined(MACOS) && !defined(ANDROID) && !defined(NOLINK_STRCASESTR)
-const char *
+char *
 strcasestr(const char *text, const char *pattern);
 #endif
 
--- a/common/utils_shared.c
+++ b/common/utils_shared.c
@@ -51,7 +51,7 @@ strnchr(const char *str, int find, size_t max)
  * want a libc dependence.
  */
 #if !defined(MACOS) && !defined(ANDROID) && !defined(NOLINK_STRCASESTR)
-const char *
+char *
 strcasestr(const char *text, const char *pattern)
 {
     const char *cur_text, *cur_pattern, *root;
@@ -60,7 +60,7 @@ strcasestr(const char *text, const char *pattern)
     cur_pattern = pattern;
     while (true) {
         if (*cur_pattern == '\0')
-            return root;
+            return (char *) root;
         if (*cur_text == '\0')
             return NULL;
         /* XXX DRi#943: toupper is better, for int18n, and we need to call


derekbruening:
@xdje42 external users seem to be hitting this issue, would be good to get this merged: https://groups.google.com/g/dynamorio-users/c/WeQqxjwuPDI

common/utils.h
```diff
@@ -1011,7 +1011,7 @@ drmem_strndup(const char *src, size_t max, heapstat_t type);
 #define MAX_OPTION_LEN DR_MAX_OPTIONS_LENGTH
```
derekbruening, line 1011: > Also, there are failures in the clang and x86 workflows, but comparing the logs before/after
shows all failures pre-exist this patch, no new failures.

This is not sufficient: as with any test failure in any pull request, you need to cite a filed issue on the failure.  If there is no issue, file a new one.

You can see these are "vsyscall incorrect assumption" which is #2491 which was caused by Github switching to AMD hardware.  Please update the PR to include this.

If you have a chance, you can contribute to cleanup by adding these to the ignore list in a separate PR to get the tests green, so the next person doesn't have to deal with red tests.

derekbruening (approved):

common/utils.h
xdje42: Have you seen the number of failures? How did it get so bad?

Issue #2548, opened 2025-03-14
Ubuntu 20.04 GA VM support being dropped April 1, 2025


derekbruening:
Github will stop supporting the Ubuntu 20.04 images on April 1.  We need to upgrade these workflows:
```
$ git grep ubuntu-2 .github/
.github/workflows/ci-aarchxx.yml:    runs-on: ubuntu-20.04
.github/workflows/ci-aarchxx.yml:    runs-on: ubuntu-20.04
.github/workflows/ci-clang.yml:    runs-on: ubuntu-20.04
.github/workflows/ci-docs.yml:    runs-on: ubuntu-20.04
.github/workflows/ci-package.yml:    runs-on: ubuntu-20.04
.github/workflows/ci-package.yml:    runs-on: ubuntu-20.04
.github/workflows/ci-x86.yml:    runs-on: ubuntu-20.04
```

There may well be test failures to fix.  Xref the work to convert the dynamorio repository workflows: https://github.com/DynamoRIO/dynamorio/issues/7270 which ran into a number of test failures.

Artem-OSSRevival:
Which version do we move to? 22.04 or 24.04?

derekbruening:
> Which version do we move to? 22.04 or 24.04?

22.04 (that is what dynamorio's tests are supporting as the link above https://github.com/DynamoRIO/dynamorio/issues/7270 shows; trying to do 24.04 may hit dynamorio problems and be a lot more work).

Artem-OSSRevival:
On Ubuntu 22.04.5 there are "_FORTIFY_SOURCE" warnings during the building.
It seems, this problem is known: https://github.com/DynamoRIO/dynamorio/issues/6606
dynamorio/ext/drsyms/CMakeLists.txt:226 has "_FORTIFY_SOURCE=3" which seems to has a meaning for the warning on the old compiler (on 24.04 there is no this warnings).
Probably, some edition of the "dynamorio" project CMakeList.txt file will be required to close both of these issues.

Artem-OSSRevival:
I consider it necessary to share the current state of affairs regarding this issue. Some problems were solved with ease, while others still remain unresolved, even after a few weeks of efforting. Perhaps you can give some clue that will help to move forward.

**ci-aarchxx.yml** solved with `android-ndk-r13b` and adding `-mfpu=vfpv3-d16` to CMakeLists.txt:331

**ci-x86.yml, ci-clang.yml**

**x64:** all tests passed

[...]

PR #2553, opened 2025-04-04
i#2548-ub22: Update DR for Ubuntu22 fixes

derekbruening:
Updates DR to 70d3287f1baeb65d4c1c9665258db0b10abdfc38 to pull in changes for building on Ubuntu22.

Issue: #2548, DynamoRIO/dynamorio#5860, DynamoRIO/dynamorio#7270


derekbruening:
Actually the failures are from the drsyscall copy:
```
CMake Error at drsyscall/CMakeLists.txt:119 (add_library):
  add_library cannot create target "drsyscall" because another target with
  the same name already exists.  The existing target is a shared library
  created in source directory
  "/home/runner/work/drmemory/drmemory/dynamorio/ext/drsyscall". 
```

@ivankyluk

ivankyluk:
There is another conflict in names:

CMake Error at CMakeLists.txt:1882 (add_executable):
  add_executable cannot create target "unit_tests" because another target
  with the same name already exists.  The existing target is an executable
  created in source directory
  "/usr/local/google/home/kyluk/DrMemory/drmemory/dynamorio/core".  See
  documentation for policy CMP0002 for more details.

derekbruening:
The unit_tests conflict is not a problem with automated testing as it never enables tests in both repos at the same time; no developer had ever done that locally before either, apparently.

Issue #1909, opened 2016-08-23
update syscall tables wrt syscalls not present on recent kernels


derekbruening:
Regular runs on win7 end up complaining about all kinds of missing system calls, most of which were removed prior to win7:

```
WARNING: could not find system call NtCancelDeviceWakeupRequest
WARNING: could not find system call NtCreateChannel
WARNING: could not find system call NtListenChannel
WARNING: could not find system call NtOpenChannel
WARNING: could not find system call NtQueryOleDirectoryFile
WARNING: could not find system call NtReplyWaitSendChannel
WARNING: could not find system call NtRequestDeviceWakeup
WARNING: could not find system call NtRequestWakeupLatency
WARNING: could not find system call NtSendWaitReplyChannel
WARNING: could not find system call NtSetContextChannel
WARNING: could not find system call NtSetHighWaitLowThread
WARNING: could not find system call NtSetLowWaitHighThread
WARNING: could not find system call NtW32Call
WARNING: could not find system call NtWow64CsrNewThread
WARNING: could not find system call NtWow64CsrSetPriorityClass
WARNING: could not find system call NtAcquireCMFViewOwnership
WARNING: could not find system call NtClearAllSavepointsTransaction
WARNING: could not find system call NtClearSavepointTransaction
WARNING: could not find system call NtListTransactions
WARNING: could not find system call NtMarshallTransaction
WARNING: could not find system call NtPullTransaction
WARNING: could not find system call NtReleaseCMFViewOwnership
WARNING: could not find system call NtRollbackSavepointTransaction
WARNING: could not find system call NtSavepointComplete
WARNING: could not find system call NtSavepointTransaction
WARNING: could not find system call NtStartTm
WARNING: could not find system call NtWow64CsrBaseClientConnectToServer
WARNING: could not find system call NtWow64CsrBasepGetTempFile
WARNING: could not find system call NtWow64CsrBasepNlsCreateSection
WARNING: could not find system call NtWow64CsrBasepNlsSetMultipleUserInfo
WARNING: could not find system call NtWow64CsrBasepNlsSetUserInfo
WARNING: could not find system call RegisterConsoleIMEInternal
WARNING: could not find system call SetConsoleCommandHistoryMode
WARNING: could not find system call UnregisterConsoleIMEInternal
WARNING: could not find system call QueryConsoleIMEInternal
WARNING: could not find system call NtUserBuildMenuItemList
[...]

PR #2556, opened 2025-06-12
Update windows syscall table with the correct min/max versions

Pastoray:
This pull request updates `drsyscall/table_windows_ntoskrnl.c` to include correct minimum and maximum Windows versions for several syscalls.

These changes were generated via a script (with the help of j00ru with the actual [x86 syscall tables](https://j00ru.vexillium.org/syscalls/nt/32/)) so their might be a minor inaccuracy or two, but overall it should be correct, this update should have no functional impact on Dr. Memory's behavior.

**Note:** Due to the current absence of specific macros in Dr. Memory that precisely define every Windows service pack or update, some entries might use broader version ranges. For instance, if you see an entry like `(WINVISTA,WINVISTA)`, please interpret this as the syscall being present from the initial release of Windows Vista (SP0) up to its final service pack (SP2). This reflects the highest achievable precision with existing definitions.

See also #1909

### DIFF
--- a/drsyscall/table_windows_ntoskrnl.c
+++ b/drsyscall/table_windows_ntoskrnl.c
@@ -174,7 +174,7 @@ syscall_info_t syscall_ntdll_info[] = {
          {10, sizeof(BOOLEAN), W|HT, DRSYS_TYPE_BOOL},
      }
     },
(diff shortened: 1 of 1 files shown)


derekbruening (approved):
Thank you for the PR!

